PR #701, opened 2020-05-14
RETAILSYS-600

vasingh:
Web Order Address Capture - NYC Borough Doesn't Populate in City Field
When Autofilled

Issue: there are addresses where google api returns city in sublocality instead of locality. In that case, City field was not being populated in the form.
examples:
175 Stockholm St (Brooklyn)

2300 Southern Boulevard (The Bronx)

2655 Richmond Ave (Staten Island)


chenson42:
@ankitvashist is this needed?

PR #725, opened 2020-06-08
Bugfix/options dialog not scrolling for large item set

blubaugha:
# Fixes
Fixed options list, in the OptionsScreenDialog component, not showing scrollbar when items overflow

Fixes bug #1107
https://github.com/JumpMind/nu-commerce/issues/1107

# Example
options-list-overflow-fix


jrmihalick:
This is not related to this issue, but I noticed that text on the buttons has recently gone from being centered to off-center.  Did that happen as a result of the key binding text changes?

chenson42:
Created #1215 for the text alignment issue.  Guessing this was due to a breaking change in Chrome like what caused the issue on the Home page buttons

PR #748, opened 2020-06-17
Edit user not scrolling Resolution

blubaugha:
# Fixes
Fixed the **Edit User** form content not scrolling when there's a lot of form fields.

Fixes [Edit Customer screen is not scrollable. #1293](https://github.com/JumpMind/commerce/issues/1293)

*Note*: Ignore the "Missing Resource" messages - I was lazy creating a test form and didn't add translations for the form field labels 😄 

# ELOPOS Resolution
edit-user-scroll-fix--elopos

# Petco Resolution
edit-user-scroll-fix--petco

# Desktop Resolution
edit-user-scroll-fix--desktop


openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/dynamic-form-dialog/dynamic-form-dialog.component.scss
```diff
@@ -3,4 +3,8 @@
 .form-body {
     display: grid;
     @extend %page-spaced-content; 
+}
+
+::ng-deep .content {
```
dankaste, line 8: You don't need to change this, but I do know that the Angular team plans to deprecate ng-deep in the future. That said we have plenty of places where this is used and we'll have to update. But if possible we should try and avoid using it to save the headache in the future.

PR #749, opened 2020-06-17
bugfix/dont-walk-simple-types

mmichalek:


chenson42:
Was this a bottleneck?

mmichalek:
Not sure if that was a real bottleneck, but I noticed we were inspecting the simple types while debugging something else.

PR #751, opened 2020-06-19
WIP #1254.  Allow multiple remote targets.

gwilmer:


openpos-service/src/main/java/org/jumpmind/pos/service/ServiceConfig.java
```diff
@@ -43,6 +45,7 @@ public void setSpecificConfig(Map<String, ServiceSpecificConfig> specificConfig)
         return specificConfig;
     }
 
+    //TODO: add this for profile
```
chenson42, line 48: Still need to add this so we can override profile settings in the database.

openpos-service/src/main/java/org/jumpmind/pos/service/strategy/RemoteOnlyStrategy.java
```diff
@@ -47,9 +57,29 @@ public String getStrategyName() {
...
     public Object invoke(ServiceSpecificConfig config, Object proxy, Method method, Map<String, Object> endpoints, Object[] args) throws Throwable {
-        int httpTimeoutInSecond = config.getHttpTimeout();
+
+        ResourceAccessException lastException=null;
+        for (String profileId : config.getProfileIds()) {
+            try {
+                return invokeProfile(serviceConfig.getProfiles().get(profileId), proxy, method, endpoints, args);
+            } catch (ResourceAccessException ex) {
+                lastException = ex;
+                logger.warn(String.format("Remote service %s unavailable.",profileId));
```
mmichalek, line 73: This should probably go ahead and log the exception.  logger.warn(String.format("Remote service %s unavailable.",profileId), ex);

PR #759, opened 2020-06-26
Feature/order pks

mmichalek:


openpos-persist/src/main/java/org/jumpmind/pos/persist/TableDef.java
```diff
@@ -24,15 +24,19 @@
...
+
     String name();
     String description() default "";
 
     /**
      * Set to true if the super class has an @TableDef that should be ignored for this module
      */
     boolean ignoreSuperTableDef() default false;
+
+    String primaryKey();
```
jrmihalick, line 41: Any particular reason you chose to make this a String instead of a String[]?

openpos-persist/src/main/java/org/jumpmind/pos/persist/impl/DatabaseSchema.java
```diff
@@ -220,17 +223,18 @@ protected void validateName(String nameToValidate, String type, String tableName
...
-        for (Field field : fields) {
-            dbTable.addColumn(createColumn(field, platform));
-        }
-    }
+//    protected void extendTable(Table dbTable, Class<?> clazz) {
+//        Field[] fields = clazz.getDeclaredFields();
+//        for (Field field : fields) {
+//            dbTable.addColumn(createColumn(field, platform));
+//        }
+//    }
```
jrmihalick, line 231: Cleanup needed on aisle 10

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
```diff
@@ -273,8 +273,10 @@ public int executeSql(String sql, Object... params) {
         }
     }
 
-    public ModelWrapper wrap(AbstractModel model){
-       return new ModelWrapper(model, databaseSchema.getModelMetaData(model.getClass()));
+    public ModelWrapper wrap(AbstractModel model) {
+        ModelWrapper wrapper = new ModelWrapper(model, databaseSchema.getModelMetaData(model.getClass()));
+        System.out.println("returning wrapper: " + wrapper);
```
jrmihalick, line 278: System.out?

openpos-persist/src/main/java/org/jumpmind/pos/persist/TableDef.java
mmichalek: Good question.  The only reason was so that we could do a constant PK with added on field, such as:
```
primaryKey = LineItemKey.LINE_ITEM_PK + "modLineSequenceNumber"
```

I couldn't figure a way to do that with arrays. Open to suggestions.

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
mmichalek: Who needs a logging framework when you've got System.out? :) 

thanks Jason.  I checked in cleanup for this.

openpos-persist/src/main/java/org/jumpmind/pos/persist/TableDef.java
jrmihalick: ```@TableDef(primaryKey=LineItemKey.LINE_ITEM_PK + "modLineSequenceNumber")``` or 
```@TableDef(primaryKey={LineItemKey.LINE_ITEM_PK + "modLineSequenceNumber", "blah blah"})``` 

doesn't work?

mmichalek: Jason and I talked this through and improved the syntax to be like:
`primaryKey = {LINE_ITEM_PK, "loyaltyPromotionId"})`

Under the hood that means that the primaryKey attribute is an array, but also still supports comma seperated string elements.

mmichalek: And TIL that you could give a single string to an annotation array, as in:
`primaryKey = "loyaltyPromotionId"`

which normally is not allwed:
image

PR #775, opened 2020-07-06
make sure the lock screen is always on top of other dialogs

dankaste:


chenson42:
Will I still be able to open dev tools?

dankaste:
> Will I still be able to open dev tools?

Yes. 
image

PR #778, opened 2020-07-07
Add support for optional failing during startup on SQL script error

jrmihalick:
@mmichalek 's idea. Both of us were recently snagged on this issue of cucumber tests not running because of a SQL script error.

- Adds `openpos.general.failStartupOnModuleLoadFailure` configuration parameter which will cause startup of the server to fail if there is any error loading the SQL scripts.  Defaults to `false` in order to preserve existing behavior.
- Will allow for us to catch failures in cucumber if a SQL script doesn't completely load. This has cost several of us troubleshooting time when tests fail because of a problem in the SQL and not in the actual tests.


dankaste:
I'm a big fan of this. I've been bitten a bunch by failing scripts

PR #790, opened 2020-07-13
update AbstractModel to add function to get the extension class by th…

stojankz:
…e className


dankaste:
Whats the use case for needing to get it by the String name of the class instead of by the type?

stojankz:
> Whats the use case for needing to get it by the String name of the class instead of by the type?

calling getExtension it in the receipt markdown file wouldn't work trying to pass in a class object.

PR #807, opened 2020-07-21
Feature/edit icon

stojankz:


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/icon.service.ts
```diff
@@ -58,6 +58,7 @@ export class IconService {
         this.icons.set('Device', ['device.svg', this.defaultIconProvider]);
         this.icons.set('Discount', ['local_offer.svg', this.defaultIconProvider]);
         this.icons.set('DollarSign', ['attach_money.svg', this.defaultIconProvider]);
+        this.icons.set('Edit', ['edit.svg', this.defaultIconProvider]);
```
dankaste, line 61: Do we still need to add it here? I thought adding a new icon will all done on the server

stojankz: If it is, I was never made aware of it. As far as I know, it doesn't work if I just add the string to ItemType, it also has to be added to icon.service.ts. If that's not the case, do you have an example you could point me to?

dankaste: by default if you just send "edit" it should ask the server to find "edit.svg" from the defaultIconProvider. It's not widely used yet but theres is also and IconConfig.java that gets populated from 'openpos.icons' in application.yml to provide a layer of abstraction from the svg file names. AbstractState and an icons() method that will retrieve it for you.

stojankz: huh did not realized that changed, thanks for the heads up

PR #812, opened 2020-07-27
Feature/item detail promotions

blubaugha:
# Changes
This updates the item details page to include available promotions

# Example

**With and without available promotions**

item-details-promos


jrmihalick:
A cucumber test would get you bonus points.  Just sayin.

PR #813, opened 2020-07-27
Exclude effectively read-only fields from form validation

jrmihalick:
- Discovered this problem when trying to edit an existing customer for Petco which had a loyalty number that was no longer valid per the current validation rules.  Even though the loyalty number fields was read only, the form would not validate and would not submit the "Update" action.  The form also did not show any errors.

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/interfaces/form-field.interface.ts
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/interfaces/form-field.interface.ts
@@ -45,6 +45,7 @@ export interface IFormElement {
     startAtDate: any;
     hintText: string;
     hideButtons: boolean;
+    readOnly: boolean;
 }
 
 export interface IDynamicListField {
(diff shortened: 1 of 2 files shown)


jrmihalick:
> Whats the difference between disabled and readOnly?

I spent some time looking at that myself and I don't think there is a logical difference, but we have uses of both spread around.  I think it needs cleaned up.

mmichalek:
@jrmihalick I think there is a difference between disabled and readOnly in how they are actually being used. Disabled is used in other places to effectively OMIT a piece of configuration completely. readOnly gives you the grey-out UI element you would expect.  Disabled would cause the UI element to disappear from the screen completely I bet.  

We probably need to think of a different word for one of these here because this has been confusing all along.

jrmihalick:
On the client side, here is one examples of how disabled is being used:

**dynamic-form-field.component.html**
```
<input *ngSwitchDefault matInput cdkFocusInitial id="{{formField.id}}" [formControlName]="formField.id" [formControl]="formGroup.controls[formField.id]" [placeholder]="getPlaceholderText(formField)"
                [formatterName]="formField.inputType" [required]="formField.required" (change)="onFormElementChanged(formField)"
                [attr.minlength]="formField.minLength" [attr.maxlength]="formField.maxLength" [attr.pattern]="formField.pattern" [readonly]="formField.disabled" [attr.tabindex]="formField.tabindex"
                [keyboardLayout]="keyboardLayout" [attr.type]="isNumericField() ? 'tel' : 'text'"
                autoSelectOnFocus>
```

mmichalek:
I see what you mean. Definitely calls for some cleanup here.

PR #821, opened 2020-08-05
Proposed fix for commerce issue 1767: Backwards compatibility issue when searching for customer transactions

jrmihalick:
- Support property name "aliases" for a given column def for case when a property name has changed
- Provides backward compatibility for older code client code to be able to query by criteria for columns/properties whose name has changed

Problem in commerce was as follows:

1. The older client sent over a `TransSearchRequest` to the `SearchReturnsTransEndpoint`.
2. The `TransSearchRequest` had `SearchCriteria` embedded within it that is referencing a column whose name has changed from trans_status to trans_status_code (See commerce issue 763).
3. In OpenPOS #658, I added a check to make sure that the parameters specified on a query can all be mapped to a specific attribute with an `@ColumnDef` annotation, otherwise an exception is thrown.  That change was designed to protect against mistakes in SearchCriteria map keys.
4. Since the `TransModel` class on the CO server no longer has an `@ColumnDef` annotation on the deprecated transStatus attribute, we get an exception on the CO server.

### DIFF
--- a/openpos-persist/src/main/java/org/jumpmind/pos/persist/ColumnDef.java
+++ b/openpos-persist/src/main/java/org/jumpmind/pos/persist/ColumnDef.java
@@ -38,4 +38,10 @@ public @interface ColumnDef {
     String description() default "";
     CrossReference[] crossReferences() default {};
     String crossReference() default "";
+    /**
+     * An optional list of alternate property names for this column that can be used when querying by criteria.
+     * Useful for providing backward compatibility for properties and columns whose names have changed.
+     * @return The list of alternate property names for the column.
+     */
+    String[] propertyAliases() default {};
 }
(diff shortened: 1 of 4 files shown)


openpos-persist/src/main/java/org/jumpmind/pos/persist/ColumnDef.java
```diff
@@ -38,4 +38,10 @@
     String description() default "";
     CrossReference[] crossReferences() default {};
     String crossReference() default "";
+    /**
+     * An optional list of alternate property names for this column that can be used when querying by criteria.
+     * Useful for providing backward compatibility for properties and columns whose names have changed.
+     * @return The list of alternate property names for the column.
+     */
+    String[] propertyAliases() default {};
```
mmichalek, line 46: any chance we could just all this aka?  or fka?  :)

jrmihalick: like "the property formerly known as prince" 

c8231756583bc5a84eb19815c3058884

PR #823, opened 2020-08-07
fix issues with processing special characters and parsing DL

dankaste:
We weren't reading the keypresses for the special characters correctly
Made the DL parsing a little more robust so we don't crash if fields are missing

### DIFF
--- a/openpos-util/src/main/java/org/jumpmind/pos/util/DriversLicense.java
+++ b/openpos-util/src/main/java/org/jumpmind/pos/util/DriversLicense.java
@@ -18,6 +18,8 @@ public class DriversLicense {
 
     private String firstLine;
 
+    private String complianceIndicator;
+
     private String fileType;
 
     private String issuerIDNumber;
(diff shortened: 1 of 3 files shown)


openpos-util/src/main/java/org/jumpmind/pos/util/DriversLicense.java
```diff
@@ -141,87 +145,99 @@ public DriversLicense(String data) {
...
+            jurisdictionVehicleClass = getString("DCA");
+            jurisdictionRestrictionCode = getString("DCB");
+            jurisdictionEndorsementCode = getString("DCD");
             expirationDate = parseDate("DBA");
-            lastName = elements.get("DCS");
-            firstName = elements.get("DAC");
-            middleNames = elements.get("DAD").split(",");
+            lastName = getString("DCS");
+            firstName = getString("DAC");
+            middleNames = getString("DAD").split(",");
```
stevencarley, line 179: Use StringUtils.split() here in case DAD is null?

PR #848, opened 2020-09-09
feature/improve-logging-during-sql-exception

mmichalek:


openpos-persist/src/main/java/org/jumpmind/pos/persist/model/SearchCriteria.java
```diff
@@ -41,6 +42,10 @@ public SearchCriteria addCriteria(String field, Object value) {
             this.criteria = new HashMap<String, Object>();
         }
 
+        if (value instanceof AbstractTypeCode) {
```
jrmihalick, line 45: You probably want to check `instanceof ITypeCode` here instead

jrmihalick (changes requested):

openpos-persist/src/main/java/org/jumpmind/pos/persist/model/SearchCriteria.java
mmichalek: Dang Jason can bounce my commits even from Yellowstone.

PR #849, opened 2020-09-09
feature/improve-logging-during-sql-exception-releases1.0

mmichalek:


mmichalek:
@gwilmer  - this should also fix the specific error you saw with the remote transaction lookup.

openpos-persist/src/main/java/org/jumpmind/pos/persist/model/SearchCriteria.java
```diff
@@ -41,6 +42,10 @@ public SearchCriteria addCriteria(String field, Object value) {
             this.criteria = new HashMap<String, Object>();
         }
 
+        if (value instanceof AbstractTypeCode) {
```
jrmihalick, line 45: You probably want to check `instanceof ITypeCode` here instead

jrmihalick (changes requested):

maxwellpettit (approved):

PR #878, opened 2020-09-23
remove the outjection of module data sources

chenson42:
this is to allow us to use spring boot's out of the box datasource which is required when accessing GCP deployed Postgres

### DIFF
--- a/openpos-devices/src/main/java/org/jumpmind/pos/devices/DevicesModule.java
+++ b/openpos-devices/src/main/java/org/jumpmind/pos/devices/DevicesModule.java
@@ -51,13 +51,7 @@ public class DevicesModule extends AbstractRDBMSModule {
         return super.securityService();
     }
 
-    @Override
-    @Bean(name = NAME + "DataSource")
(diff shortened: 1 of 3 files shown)


dankaste (approved):

mmichalek:
Does this remove the possibility of using our logging driver?

chenson42:
> Does this remove the possibility of using our logging driver?

only if you are using the spring datasoure

mmichalek:
So right now we have 1) SQL debug logging 2) Long Running SQL Logging 3) And logging for super long in-process SQL - built into the driver.  It would be a bummer for AEO to just not have any of that. Full disclosure: I am not a fan of container-managed datasources for reasons like this. 
image

chenson42:
it is spring managed.  i haven't gotten out OOB datasources to work with GCP postgres.   spent a day just getting the app talking to GCP postrges a few weeks back.   I can try to wrap it with the openpos driver.  guessing spring boot instrumentation will allow us to get those stats as well.  i think we should be leveraging spring boot for all instrumentation.

PR #879, opened 2020-09-23
Feature/add logging for scanner registration

craigsaw:
Added some client-side logging around scanner registration that was helpful while trying to debug a related issue (#2106)


mmichalek (approved):

dankaste:
we probably don't need this anymore. Not sure we want the extra logging as console logs are expensive

PR #880, opened 2020-09-24
Added transaction search noResultsMessage

tsarenkoa:


openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/TransactionSearchUIMessage.java
```diff
@@ -12,6 +12,7 @@
     private ActionItem clearButton;
 
     private String providerKey;
+    private String noResultsMessage;
```
chenson42, line 15: While in here we should transition this to use lombok's @Data annotation and remove the boiler plate code

tsarenkoa: done

chenson42 (approved):

stojankz (approved):

openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/TransactionSearchUIMessage.java
stevencarley: The Data annotation doesn't add a no-arg constructor. Probably need to add @NoArgsConstructor to avoid any Jackson Serialization issues.

tsarenkoa: [TransactionSearchUIMessage](https://github.com/JumpMind/openpos-framework/blob/44e348f79df63f5ab181f64b3dbcecd77de9cb62/openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/TransactionSearchUIMessage.java#L16) already contains a no-arg constructor

stevencarley: Missed that, thanks.

stevencarley (approved):

PR #891, opened 2020-10-02
Bugfix/status bar

liambell-le:


dankaste:
Can you put some details in the description?

PR #897, opened 2020-10-05
Add ability to override states

maxwellpettit:
## Summary
Add an annotation interface that allows you to override a State.  For example, using the code below, any instance of the StateToOverride state in the yml flows would be map to the OverrideState class instead:

```
    // This state will not get used
    public static class StateToOverride {
        @OnArrive
        public void arrive(){
        }
    }

    // This state will be used in place of StateToOverride in all flows
    @StateOverride(originalState = StateToOverride.class)
    public static class OverrideState {
        @OnArrive
        public void arrive(){
        }
    }
```

### DIFF
--- a/openpos-flow/src/main/java/org/jumpmind/pos/core/flow/config/YamlConfigConverter.java
+++ b/openpos-flow/src/main/java/org/jumpmind/pos/core/flow/config/YamlConfigConverter.java
@@ -11,13 +11,15 @@ import org.slf4j.LoggerFactory;
 
 
 public class YamlConfigConverter {
-    
+
     private static final Logger log = LoggerFactory.getLogger(YamlConfigConverter.class);
-    
+
     private static Map<String, Class<Object>> knownFlowClasses;
 
     private static Map<String, Class<Object>> knownOnEventClasses;
-    
+
+    private static Map<String, Class<Object>> overrideClassMap;
+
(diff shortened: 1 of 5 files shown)


openpos-flow/src/main/java/org/jumpmind/pos/core/flow/config/YamlConfigConverter.java
```diff
@@ -194,31 +196,51 @@ protected YamlFlowConfig findFlowByName(String stateName, List<YamlFlowConfig> y
...
                     .filter(clazz -> filterStateClass(clazz, allowGlobalActionHandler))
-                    .collect(Collectors.toMap(e -> ((Class)e).getSimpleName(), v -> v) );
+                    .collect(Collectors.toMap(e -> ((Class) e).getSimpleName(), v -> v));
+        }
+
+        updateOverrideClasses();
+
+        Class<? extends Object> stateClass = knownFlowClasses.get(name);
+        if (overrideClassMap.get(name) != null) {
+            stateClass = overrideClassMap.get(name);
```
mmichalek, line 215: One thing - can we log at info level when a state override occurs?  I think these will only get logged once during startup.

chenson42 (approved):

dankaste (approved):

blubaugha (approved):

PR #903, opened 2020-10-07
Improve the management of sql & csv scripts

chenson42:
### Summary
We now support running sql scripts again if they have changed.  The name of the file and the md5 checksum is stored in a {module_prefix}_script_version table.  If a file has changed (its checksum changed) then it is run again.   This means that if files change they need to be rerunnable.   For CSV files, if they have changed then the table is truncated and reloaded.

### Parameter Changes
- The `openpos.general.sqlScriptProfile` parameter has been deprecated.

### Database Changes
- A new table was added that tracks files that have been run. `{module_prefix}_script_version`

### DIFF
--- a/openpos-client-libs/package-lock.json
+++ b/openpos-client-libs/package-lock.json
@@ -2070,7 +2070,8 @@
     "boolean": {
       "version": "3.0.1",
       "resolved": "https://registry.npmjs.org/boolean/-/boolean-3.0.1.tgz",
-      "integrity": "sha512-HRZPIjPcbwAVQvOTxR4YE3o8Xs98NqbbL1iEZDCz7CL8ql0Lt5iOyJFxfnAB0oFs8Oh02F/lLlg30Mexv46LjA=="
+      "integrity": "sha512-HRZPIjPcbwAVQvOTxR4YE3o8Xs98NqbbL1iEZDCz7CL8ql0Lt5iOyJFxfnAB0oFs8Oh02F/lLlg30Mexv46LjA==",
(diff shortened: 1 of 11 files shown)


mmichalek:
I like it, this will help a lot.  I am nervous about the "truncate" part of the CSV.  Seems like that could happen in production at the wrong time.

chenson42:
> I like it, this will help a lot. I am nervous about the "truncate" part of the CSV. Seems like that could happen in production at the wrong time.

Maybe I should make that configurable?

mmichalek:
Yeah that might help.  There was also this idea from Petco where when we go to prod central, we might want commerce to generate DLL scripts instead of directly applying the DLL changes to the central DB.  So maybe this could work in conjunction with that config where the system creates a package of scripts to run instead of directly applying them.

mmichalek:
Another thought for the CSV: could we add support for SQL commands in there? So the CSV's could be written to run safely repeatable like the other .sql scripts. Like:
`SQL_CMD,delete from blah where foo='bar'`

jrmihalick:
> > I like it, this will help a lot. I am nervous about the "truncate" part of the CSV. Seems like that could happen in production at the wrong time.
> 
> Maybe I should make that configurable?

You'd probably want to also have the ability to control what specific entities are ok to truncate and which aren't, with the default being that none are truncated.  I think it would also make sense to have all of this configuration in it's own yaml file, for easier review/update.

dankaste (approved):

stevencarley (approved):

PR #905, opened 2020-10-09
Keybinds Stop Working

blubaugha:
### Summary
Fixes issue https://github.com/JumpMind/commerce/issues/2197, where the keyboard stopped working after removing a coupon. 

This bug turned out to be a general bug in the application. It's caused by clicking an element that gets removed from the DOM, which results in no content having focus so the `appKeypressSource` never receives any key press events.

The fix works by listening for DOM changes and, if there is no element with focus, then focus is brought back to the application.

### Screenshots
Here's an example of the keybindings working after clicking an element that's removed from the DOM. This scenario results in no element in the `document.body` having focus so the browser sets focus to the `document.body`.

keybinds-stop-working


dankaste (approved):
Awesome find and fix. I think we've been circling around this issue for at least a year if not longer.

chenson42:
Don't forget to make a pull request for master.

chenson42 (approved):

stevencarley (approved):

stojankz (approved):

PR #908, opened 2020-10-10
feature/print-effecient-code-128-barcode

mmichalek:
This change uses CODEC CODE 128 encoding for numeric barcodes, and handles switching to CODEB if there is an odd number of digits (in order to print the last digit in CODEB).  The result is we can set the barcode bars twice as wide.  The "after" barcode was much easier to scan on my phone - I could barely scan the first one.

image

image


jrmihalick (approved):

jrmihalick:
The "after" receipt looks much better.

jrmihalick (approved):

PR #928, opened 2020-10-19
PCOM-191: Update of collapsed Sell Item design and Image service mechanism

avramovicae:
- updated SellItem interface with collapsed labels
- updated workflow of Image service and Image Component mechanism so that
  we now cover use case scenario when Alt image property is misssing or is
  failing.
- updated ItemCard and MobileItem item collapsed looks
- updated SellItem object with collapsedAdditionalLabels property

### Commerce dependencies 
https://github.com/JumpMind/commerce/pull/2376

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/image.service.ts
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/image.service.ts
@@ -41,7 +41,7 @@ export class ImageService {
     setImageNotFoundURL(){
         this.session.getMessages('ConfigChanged')
             .pipe(
-                filter(message => message.configType === 'MediaService'),
+                filter(message => message.configType === 'ImageService'),
                 map(res => res["image-not-found"] ),
                 take(1)
             ).subscribe(res => this._imageNotFoundURL = res);
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/image.service.ts
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/image.service.ts
@@ -2,6 +2,7 @@ import {Injectable} from '@angular/core';
 import {SessionService} from './session.service';
 import {DiscoveryService} from '../discovery/discovery.service';
 import {PersonalizationService} from '../personalization/personalization.service';
+import {filter, map, take} from "rxjs/operators";
 
 @Injectable({
     providedIn: 'root',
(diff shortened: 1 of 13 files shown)


chenson42:
@avramovicae Can we get a few screen shots of this in action?

avramovicae:
@chenson42 Here are screenshots. For the first two I moved style from list of collapsed labels so that there is more visible difference between collapsed and expanded item, since at the moment additional labels by default only contain three properties that are also in collapsed Additional labels. In the last one it shows current state. 
Note: We should check with AE should there be adjustments in default additional labels in SellItem.

Screenshot 2020-10-20 at 12 36 18
Screenshot 2020-10-20 at 12 36 09

Screenshot 2020-10-20 at 12 46 32

avramovicae:
@dankaste @chenson42 I have additionally updated image component logic.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/image.service.ts
```diff
@@ -10,24 +11,35 @@ export class ImageService {
...
+        let url = originalUrl.replace(this.baseUrlToken, apiServerBaseUrl);
+        url = url.replace(this.appIdToken, appId);
+        url = url.replace(this.deviceIdToken, deviceId);
+        return url;
+    }
 
+    setImageNotFoundUrl(){
+        this.session.getMessages('ConfigChanged')
+            .pipe(
+                filter(message => message.configType === 'MediaService'),
```
dankaste, line 39: This should be getting its configuration from a different config set than 'MediaService'. I was picturing a config set for 'Images'.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/image.service.ts
```diff
@@ -10,24 +11,35 @@ export class ImageService {
...
-            const deviceId = this.personalizer.getDeviceId$().getValue();
-            const appId = this.personalizer.getAppId$().getValue();
-
-            let url = originalUrl.replace(this.baseUrlToken, apiServerBaseUrl);
-            url = url.replace(this.appIdToken, appId);
-            url = url.replace(this.deviceIdToken, deviceId);
-            return url;
-        } else {
-            return originalUrl;
+        if(!originalUrl){
```
dankaste, line 23: I don't think it really makes sense to do this here cause you don't know if the original image has failed to fetch yet. Also this doesn't really fallow the single responsibility principle. This method should only be replacing the tokens in the url and returning back the correct path.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/image/image.component.ts
```diff
@@ -29,16 +30,38 @@ export class ImageComponent {
...
-      image.src = altImage;
+    let imgSource = this.imageSources[0];
+    let imgUrl;
+    switch (imgSource) {
+      case ImageSource.ALT_IMG:
+        this.orgImgSrc = image.src;
+        imgUrl = this.imageService.replaceImageUrl(this.altImageUrl);
+        break;
+      case ImageSource.DEFAULT_NOT_FOUND_IMG:
+        imgUrl = this.imageService.replaceImageUrl('');
```
dankaste, line 54: Instead of passing in blank, I think its much more clear if we pass in imageServce.getDefaultImage()

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/mobile-item/mobile-item.component.ts
```diff
@@ -11,8 +11,12 @@ import { ActionService } from '../../../core/actions/action.service';
     styleUrls: ['./mobile-item.component.scss']
 })
 export class MobileItemComponent {
-
-    @Input() item: ISellItem;
+    item: ISellItem;
+    filteredAdditionalLabels:{label: string, value: string} [];
+    @Input() set _item(item: ISellItem){
```
dankaste, line 16: I'm not crazy about the '_' prefix on the public property. I believe the best practice is the private has the underscore and the property is camelCase

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/item-card/item-card.component.ts
```diff
@@ -17,7 +17,13 @@ import { KebabLabelButtonComponent } from '../kebab-label-button/kebab-label-but
 })
 export class ItemCardComponent implements OnDestroy {
 
-  @Input() item: ISellItem;
+  item:ISellItem;
+  filteredAdditionalLabels:{label: string, value: string} [];
+
+  @Input() set _item(item: ISellItem){
```
dankaste, line 23: Swap the underscore prefix with the private variable.

dankaste (changes requested):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/image.service.ts
avramovicae: @dankaste I already created PR with image-not-found in MediaService in yml file in aeo-jumpmind, and this PR is approved and merged by @chenson42 . Should I update this yml file or we can go with this?

dankaste: Yes please update the yaml file.

dankaste (changes requested):
Just need to get the ConfigChanged message from the right place

stevencarley (approved):

stojankz (approved):

dankaste (approved):

PR #929, opened 2020-10-19
TagProviders are not being called unless you have tags configured already (in which case you wouldn't need the provider)

chenson42:
### Summary
TagProviders are not being called unless you have tags configured already (in which case you wouldn't need the provider)


jrmihalick (commented):
Does this mean that properties like this currently in the application-env.properties:

`openpos.tags.STATE=NY`

need to be changed to

`openpos.tagconfig.tags.STATE=NY`

?

chenson42:
> Does this mean that properties like this currently in the application-env.properties:
> 
> 
> 
> `openpos.tags.STATE=NY`
> 
> 
> 
> need to be changed to
> 
> 
> 
> `openpos.tagconfig.tags.STATE=NY`
> 
> 
> 
> ?

No.  The calling of the provider api is now being driven by config versus actual values.

jrmihalick (approved):

mmichalek (approved):

PR #934, opened 2020-10-26
Refactor IMessageInterceptor logic to allow for ordered interception

jrmihalick:
### Issues Fixed
- Adds support needed for PDPOS-3950

### Summary
- There was no way to add IMessageInterceptors from client code and control the order in which the interceptor was executed.  This change allows for IMessageInterceptors to be executed in the order defined by their `order` method, which enables client code to supplant behavior in the core commerce code if needed.
- Also adding some logging in the lowest level API for opening the cash drawer, so that we ensure we get a log message output whenever commerce code triggers a cash drawer to open.

### DIFF
--- a/openpos-peripheral/src/main/java/org/jumpmind/pos/print/EscpPOSPrinter.java
+++ b/openpos-peripheral/src/main/java/org/jumpmind/pos/print/EscpPOSPrinter.java
@@ -8,7 +8,6 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.text.StrSubstitutor;
 import org.jumpmind.pos.util.ClassUtils;
 import org.jumpmind.pos.util.status.Status;
-import org.jumpmind.pos.util.status.StatusReport;
 
 import javax.imageio.ImageIO;
 import java.awt.image.BufferedImage;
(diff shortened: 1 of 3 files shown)


openpos-peripheral/src/main/java/org/jumpmind/pos/print/EscpPOSPrinter.java
```diff
@@ -102,6 +104,10 @@ public void printNormal(int station, String data) {
                     throw new PrintException("The output stream for the printer driver cannot be null " +
                             "at this point. It probably was not initialized properly. (Hint: you may need to call open()");
                 }
+                if (StringUtils.containsIgnoreCase(data, printerCommands.get(PrinterCommands.CASH_DRAWER_OPEN))) {
```
mmichalek, line 107: I think you can just do "contains" vs. containsIgnoreCase.   The open drawer command is just a series of bytes, there isn't a notion of "case" like you have with characters.  What do you think?

jrmihalick: Sure, you know this data better than I.  It was a CYA line of code.  :-)

chenson42:
I typically just use the spring annoation @Order to control the order of a list of spring objects

mmichalek (approved):

jrmihalick:
> I typically just use the spring annoation @order to control the order of a list of spring objects

@chenson42 Yeah, I agree and that is the first option I tried, but it doesn't work in this case due to the way that the interceptors are getting loaded.  See the `ScreenService.interceptMessage(String, String, T, Class<T>)` method on the "Files Changed" tab of this PR.

PR #936, opened 2020-10-26
Bugfix/printer crash fix

mmichalek:
Port update to jserialComm


chenson42 (dismissed):

openpos-peripheral/src/main/java/org/jumpmind/pos/print/EscpCashDrawerService.java
```diff
@@ -0,0 +1,312 @@
...
+            e.printStackTrace();
+        }
+    }
+
+    public void openDrawer() throws JposException {
+        init();
+        if (printer != null) {
+            try {
+                openPrinter();
+                printer.openCashDrawer("");
```
jrmihalick, line 111: Log a message here?

jrmihalick (approved):

PR #937, opened 2020-10-27
JMC Demo Enhancements

liambell-le:
### Summary
Local Storage has been moved to first tab in Dev Tools and changed to "Options". 
If a simURL is not undefined for the simulated peripheral (receipt printer), the simUrl will be used in lieu of a different port on the same hostname.


openpos-flow/src/main/java/org/jumpmind/pos/core/service/DevToolsActionListener.java
```diff
@@ -51,11 +51,17 @@
...
 
-    @Value("${server.secondary.ports:7400}")
-    String ports[];
+    @Value("${openpos.peripheralSimulatorViewer.simPort}")
+    String simPort;
 
     @Value("${openpos.peripheralSimulatorViewer.appId:'sim'}")
     String simAppId;
+
+    @Value("${openpos.peripheralSimulatorViewer.simUrl}")
```
stevencarley, line 60: Set a default of null if these values aren't set in the configuration.

stevencarley (changes requested):

openpos-flow/src/main/java/org/jumpmind/pos/core/service/DevToolsActionListener.java
liambell-le: Done

stevencarley (approved):

PR #941, opened 2020-11-01
Cancel all executors after the state manager has been reset

chenson42:
### Summary
State's can start asynchronous threads via `AsyncExecutors`.  If someone logs out while an AsyncExecutor is running then it needs to be marked as cancelled so that the results are handled appropriately.

### DIFF
--- a/openpos-flow/src/main/java/org/jumpmind/pos/core/flow/AsyncExecutor.java
+++ b/openpos-flow/src/main/java/org/jumpmind/pos/core/flow/AsyncExecutor.java
@@ -27,6 +27,8 @@ public class AsyncExecutor {
     @Autowired
     StateManagerContainer stateManagerContainer;
 
+    Runnable beforeCancel;
+
     @PostConstruct
     public void init() {
         this.scheduler = new ThreadPoolTaskScheduler();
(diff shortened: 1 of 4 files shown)


mmichalek (approved):

dankaste (approved):

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/AsyncExecutor.java
```diff
@@ -33,14 +38,35 @@ public void init() {
...
-    public <T, R> void execute(T request, Function<T, R> doWork, Consumer<R> handleResult, Consumer<Throwable> handleError) {
+    synchronized public void cancel() {
+        cancelled = true;
+        if (beforeCancel != null) {
+            this.beforeCancel.run();
+        }
+    }
+
+    synchronized public <T, R> void execute(T request, Function<T, R> doWork, Consumer<R> handleResult, Consumer<Throwable> handleError) {
+       execute(request, doWork, handleResult, handleError, r-> log.info("The executor returned but the results are being ignore because it was cancelled"), null);
```
jrmihalick, line 49: ignore -> ignored

jrmihalick (changes requested):

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/AsyncExecutor.java
mmichalek: Maybe in that vein we could log a hint as to what executor is being ignored, or what results are being discarded?

jrmihalick (approved):

maxwellpettit (approved):

PR #948, opened 2020-11-03
Bugfix/keypress sources capturing both keyup and keydown

stojankz:
### Summary
Change fixes an issue where pressing 'Enter' to unlock the lock screen was also using the enter keypress to send the open store dialog action and continue opening the store. This issue resulted from the fact that both keyup and keydown events were being handled for the keypress sources which resulted in the lock screen being handled on keydown and the dialog underneath to capture and use keyup.

Changes made include:
- updating the lock-screen component to use the KeyPressProvider and keypress-source to handle the keypress rather than using a host listener
- updating keypress-source.directive.ts to make sure it's only registering the source to the provider and not also subscribing to keyup and keydown events in the source directive
- updating keypress.provider.ts to add additional optional parameter to subscribe to specify the eventType that the key should subscribe to. If this parameter is not passed in, it subscribes to 'keydown' by default.
- updating keypress.provider.ts rebuildKeyPressObserver function to filter on the event type of the subscriptions and then sort so that it is handling only subscriptions for that key and that eventType(keyup or keydown).
- updating keypress.provider.ts to check the Configuration.keybindsEnabled in the shouldRunGlobalAction function
- updating action-item-key-mapping.directive.ts so that it subscribes to both keyup and keydown to continue to handle the button styling
- cleaned up spacing and formatting


chenson42 (approved):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/providers/keypress.provider.ts
```diff
@@ -63,14 +63,15 @@ export class KeyPressProvider implements OnDestroy {
     }
 
     shouldRunGlobalAction(action: IActionItem): boolean {
+        // do we need to check if lock screen is enabled now that we stop propagation?
```
blubaugha, line 66: I think you're right - you don't need this check anymore

blubaugha (approved):

PR #958, opened 2020-11-07
Process events on the action loop in order to eliminate event and action processing race conditions

chenson42:
### Summary
Process events on the action loop in order to eliminate event and action processing race conditions.  Simultaneous processing of events and actions has caused sequence numbers to be skipped in the past, and this is the solve for that.


chenson42:
Much more elegant than the previous attempt at solving this issue.

jrmihalick (approved):
Looks great

stevencarley (approved):

dankaste (approved):

PR #960, opened 2020-11-09
Auto-focus on first options element

JJStarrett:
### Issues Fixed
 - [4103](https://petcoalm.atlassian.net/browse/PDPOS-4103)

### Summary
The first element of each options list can now optionally be auto-focused.

### Screenshots
23cd2b4ecc5c6fca89a70bee24b9eec4

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/icon-button/icon-button.component.html
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/icon-button/icon-button.component.html
@@ -1,8 +1,7 @@
-<button responsive-class #button mat-icon-button [ngClass]="iconClass"
+<button mat-icon-button [ngClass]="iconClass"
     class="icon-button"
     [disabled]="disabled"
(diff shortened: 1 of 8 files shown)


stojankz (approved):

chenson42 (approved):

chenson42:
I wonder if this should be configurable from the server ui message?

JJStarrett:
> I wonder if this should be configurable from the server ui message?

I'm not completely sure what the server ui message is, but I'd be happy to implement it.

chenson42:
I was thinking maybe on OptionsListPart.java you add a setting as to whether the first item gets active focus or not.   I can imagine there are times when we'll want to the user to explicitly choose something.

dankaste (approved):

JJStarrett:
> I was thinking maybe on OptionsListPart.java you add a setting as to whether the first item gets active focus or not. I can imagine there are times when we'll want to the user to explicitly choose something.

Added.

chenson42 (approved):

PR #998, opened 2020-11-27
PCOM-668: Integrate web sdk for scandit scanner

avramovicae:
- created component for scandit scanner view
- created plugin file that handles validation with key and scan actions
- updated core module with these files references


kdelorey:
I did some exploring of this. It doesn't seem feasible to use the Web SDK from Scandit on iOS. Directly in Safari requires the page to be served from a TLS encrypted endpoint (https) and if you try to run it within WebKit apps (Cordova/Capacitor) it will not grant you access to any video streams. This limits the Web SDK to only be used with "Single Image Mode" which requires the application to open the native camera app, take a photo, and then share that photo with the web page.

PR #1007, opened 2020-12-03
add getters to model classes for a session factory

dankaste:
### Summary
Add getters to the model classes so we can see what models are associated with a modules SessionFactory

### DIFF
--- a/openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSessionFactory.java
+++ b/openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSessionFactory.java
@@ -31,7 +31,7 @@ public class DBSessionFactory {
     IDatabasePlatform databasePlatform;
     TypedProperties sessionContext;
     @Getter
-    List<Class<?>> modelClazzes;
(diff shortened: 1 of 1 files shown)


openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSessionFactory.java
```diff
@@ -29,7 +30,9 @@
     Map<String, DmlTemplate> dmlTemplates;
     IDatabasePlatform databasePlatform;
     TypedProperties sessionContext;
+    @Getter
```
mmichalek, line 33: Can we correct the spelling of 'clazzes', especially now that it's getting publicized?

stevencarley (approved):

chenson42 (approved):

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSessionFactory.java
dankaste: whats wrong with clazzes?

PR #1008, opened 2020-12-03
Mask scan data to hide info like drivers licenses

maxwellpettit:
### Summary
Mask scan data to hide sensitive info like drivers licenses when the server receives an Action

The resulting logging for the action would look like:
```
2020-12-03 16:41:52.818 INFO  [pos:11111-111] [clientInboundChannel-2] [ScreenService] Received action from 11111-111
{
  "name" : "ScanDriversLicense",
  "data" : {
    "data" : "********************",
    "rawData" : "********************"
  },
  "type" : "Screen",
} 
```

### DIFF
--- a/openpos-flow/src/main/java/org/jumpmind/pos/core/util/ActionSerializer.java
+++ b/openpos-flow/src/main/java/org/jumpmind/pos/core/util/ActionSerializer.java
@@ -1,12 +1,10 @@
 package org.jumpmind.pos.core.util;
 
 import java.io.IOException;
+import java.util.Map;
 
 import org.apache.commons.lang3.StringUtils;
-import org.jumpmind.pos.core.model.FieldInputType;
-import org.jumpmind.pos.core.model.Form;
-import org.jumpmind.pos.core.model.FormField;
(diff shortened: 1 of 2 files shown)


chenson42 (approved):

mmichalek:
Hmm.. This is tough because the vast about of scans will not be sensitive and we lose a lot of logging intelligence not knowing what the user scanned at all anymore.. Any chance the masking could be pattern matched?

maxwellpettit:
> Hmm.. This is tough because the vast about of scans will not be sensitive and we lose a lot of logging intelligence not knowing what the user scanned at all anymore.. Any chance the masking could be pattern matched?



> Hmm.. This is tough because the vast about of scans will not be sensitive and we lose a lot of logging intelligence not knowing what the user scanned at all anymore.. Any chance the masking could be pattern matched?

I made a change so that "Scan" is no longer a sensitive action name.  The caveat here was that I also had to go to specific screens where we would be scanning the DL, and manually set the action name to "ScanDriversLicense" rather than just the default "Scan".  Luckily there's only a couple of places I had to do this, so not a huge change.

mmichalek (approved):

PR #1009, opened 2020-12-04
Toast container

eddielegg-le:
### Summary
Change implementation of toast messages to support business date mismatch message:
- Changed library for messages to support multiple messages
  - Required multiple changes, including style changes and implementation of custom toast component
- Updated Toast class to accept an icon
- Added `info` color scheme

### Screenshots
Samples of toast messages
image

Business Date Message
image

### DIFF
--- a/openpos-flow/src/main/java/org/jumpmind/pos/core/ui/Toast.java
+++ b/openpos-flow/src/main/java/org/jumpmind/pos/core/ui/Toast.java
@@ -11,6 +11,7 @@ public class Toast extends Message {
     private ToastType toastType;
     private int duration = 2500;
     private String verticalPosition = "bottom";
+    private String icon;
 
     public static Toast createSuccessToast(String message) {
         return createSuccessToast(message, true);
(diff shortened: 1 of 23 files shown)


blubaugha (approved):

openpos-flow/src/main/java/org/jumpmind/pos/core/ui/Toast.java
```diff
@@ -11,6 +13,10 @@
     private ToastType toastType;
     private int duration = 2500;
     private String verticalPosition = "bottom";
+    private String icon;
+    private boolean persistent;
+    private UUID persistedId;
+    private boolean close;
```
eddielegg-le, line 19: Think maybe this should be a different message? Maybe a `CloseToast` class with just the ID?

mmichalek:
Eddie this is looking good.  Any chance we can widen that toast, so the business date one is one line and allows the logo to peek through the back.
image

mmichalek (commented):
See widen comment.

eddielegg-le:
> Eddie this is looking good. Any chance we can widen that toast, so the business date one is one line and allows the logo to peek through the back.

Sure can:

image

blubaugha (approved):

mmichalek (approved):

PR #1010, opened 2020-12-07
Support deviceless applications on the openpos framework.  A central office dashboard would be an example.

chenson42:
### Summary
Support deviceless applications on the openpos framework.  A central office dashboard would be an example.

### Parameter Changes
Set openpos.services.specificConfig.devices.implementation to the value deviceless and set openpos.general.defaultAppId to the appId of the application being served to enable an application that does not have to be personalized.

### DIFF
--- a/openpos-devices/src/main/java/org/jumpmind/pos/devices/model/DevicelessRepository.java
+++ b/openpos-devices/src/main/java/org/jumpmind/pos/devices/model/DevicelessRepository.java
@@ -1,0 +1,36 @@
+package org.jumpmind.pos.devices.model;
+
+import org.springframework.stereotype.Component;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@Component
+public class DevicelessRepository {
+    Map<String, DeviceModel> virtualDevices = new HashMap<>();
+
+    public void add(String authToken, DeviceModel deviceModel) {
+        virtualDevices.put(authToken, deviceModel);
+    }
+
+    public DeviceModel getByAuthToken(String authToken) {
+        return virtualDevices.get(authToken);
+    }
+
+    public DeviceModel getByDeviceIdAppId(String deviceId, String appId) {
+        return virtualDevices.values().stream().filter(d->d.getDeviceId().equals(deviceId) && d.getAppId().equals(appId)).findFirst().orElse(null);
+    }
+
+    public void removeByDeviceIdAppId(String deviceId, String appId) {
+        String matchingAuthId = null;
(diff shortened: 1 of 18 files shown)


dankaste (approved):

stojankz (approved):

stevencarley (approved):

openpos-devices/src/main/java/org/jumpmind/pos/devices/model/DevicelessRepository.java
```diff
@@ -0,0 +1,36 @@
+package org.jumpmind.pos.devices.model;
+
+import org.springframework.stereotype.Component;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@Component
+public class DevicelessRepository {
+    Map<String, DeviceModel> virtualDevices = new HashMap<>();
```
mmichalek, line 10: This probably needs thread safety.  ConcurrentHashMap?

openpos-devices/src/main/java/org/jumpmind/pos/devices/service/GetMyDevicelessEndpoint.java
```diff
@@ -0,0 +1,26 @@
+package org.jumpmind.pos.devices.service;
+
+import org.jumpmind.pos.devices.service.model.GetDeviceRequest;
+import org.jumpmind.pos.devices.service.model.GetDeviceResponse;
+import org.jumpmind.pos.service.Endpoint;
+import org.jumpmind.pos.util.clientcontext.ClientContext;
+import org.springframework.beans.factory.annotation.Autowired;
+
+@Endpoint(path = "/devices/myDevice", implementation = "deviceless")
+public class GetMyDevicelessEndpoint {
```
mmichalek, line 10: This could be a puzzling name in the future.

mmichalek:
I wonder if there is a different description we could use in place of "deviceless"?  I read somewhere that negative names in code require a bit more thinking to understand.  Also, I'm not sure that deviceless is wholly accurate.  There usually is a device involved, it's just not a registered, corporate device I guess.

What do you think?  Some ideas:
ViritualDevice, FloatingDevice, PublicDevice, WebDevice, SharedDevice?  Just brainstorming.

jrmihalick:
I agree and thought the same thing about it being confusing, but at the time I couldn't think of anything better to suggest.  To me, there is a device, but it's anonymous or a corp device.  Corp device?

mmichalek (approved):

PR #1012, opened 2020-12-10
Updates for core app dashboard screen

eddielegg-le:
### Summary
Create components to support dashboard screen:
- Added DashboardUIMessage class and added UIMessageType


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/personalization/personalization.component.ts
```diff
@@ -163,7 +163,7 @@ export class PersonalizationComponent implements IScreen, OnInit {
 
     public personalizeLocal(): Observable<string> {
         const personalizationProperties = new Map<string, string>();
-        if (this.serverResponse) {
+        if (this.serverResponse && this.serverResponse.parameters) {
```
eddielegg-le, line 166: Not sure what state I was in to cause this, but I had a `serverResponse` but `parameters` was null

blubaugha (approved):

PR #1017, opened 2020-12-14
Retain focus on poptart input

eddielegg-le:
### Issues Fixed
https://petcoalm.atlassian.net/browse/PDPOS-4496

### Summary
PopTart input types were losing focus when changed, causing the user to have to manually advance the form rather than just hitting the enter button. This change ensures the input retains focus when the dialog has been closed.


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/dynamic-form-field/dynamic-form-field.component.ts
```diff
@@ -180,6 +180,7 @@ export class DynamicFormFieldComponent implements OnInit, OnDestroy, AfterViewIn
     dialogRef.afterClosed().subscribe(result => {
       console.info('pop tart closed with value of: ' + result);
       this.formGroup.get(this.formField.id).setValue(result);
+      this.field.focus();
```
dankaste, line 183: Material Dialogs have a setting in the properties when you open them to restore focus when they close. I honestly don't know if its functionally any different than what you are doing here but for the sake of knowledge ... https://material.angular.io/components/dialog/api

dankaste (approved):

PR #1019, opened 2020-12-15
Retain focus on poptart input (#1017)

eddielegg-le:
### Summary
Cherry pick into master: retain focus on poptart input after dialog closes


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/dynamic-form-field/dynamic-form-field.component.ts
```diff
@@ -180,6 +180,7 @@ export class DynamicFormFieldComponent implements OnInit, OnDestroy, AfterViewIn
     dialogRef.afterClosed().subscribe(result => {
       console.info('pop tart closed with value of: ' + result);
       this.formGroup.get(this.formField.id).setValue(result);
+      this.field.focus();
```
dankaste, line 183: You should also be able to set the DialogProperty for Restore focus when we open the poptart.

dankaste (approved):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/dynamic-form-field/dynamic-form-field.component.ts
eddielegg-le: I put this in there to handle the case when the button was clicked to open the dialog instead of the input itself (as well as clicking the input)

PR #1040, opened 2020-12-27
Log calls to all endpoints with the exception of endpoints tagged to suppress logging

chenson42:
### Summary
Log calls to all endpoints with the exception of endpoints tagged to suppress logging.


chenson42:
Renamed EndpointDispactchInvocationHandler to EndpointInvoker just because it shows up in the log files quite often now.  I liked seeing the shorter name.

mmichalek (approved):

PR #1059, opened 2021-01-07
Mask scan data to hide info like drivers licenses (#1008)

maxwellpettit:
### Summary
(Merge from 1.1) Mask scan data to hide sensitive info like drivers licenses when the server receives an Action

The resulting logging for the action would look like:
```
2020-12-03 16:41:52.818 INFO  [pos:11111-111] [clientInboundChannel-2] [ScreenService] Received action from 11111-111
{
  "name" : "ScanDriversLicense",
  "data" : {
    "data" : "********************",
    "rawData" : "********************"
  },
  "type" : "Screen",
} 
```


maxwellpettit:
This apparently did not get merged from 1.1 back in December, but it's a requirement to mask PII data in logs, specifically driver's licenses.

jrmihalick (approved):

PR #1062, opened 2021-01-08
Allow java.util.Date and other java objects to be queries from the persistence layer

chenson42:
### Summary
Allow java.util.Date and other java objects to be queries from the persistence layer

### DIFF
--- a/openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
+++ b/openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
@@ -1,21 +1,5 @@
 package org.jumpmind.pos.persist;
 
-import static org.apache.commons.lang3.StringUtils.isNotBlank;
-
-import java.beans.PropertyDescriptor;
-import java.io.File;
-import java.io.FileReader;
-import java.io.Reader;
-import java.lang.reflect.Field;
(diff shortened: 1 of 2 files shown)


jrmihalick (approved):

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
```diff
@@ -662,6 +661,11 @@ public void batchUpdate(List<? extends AbstractModel> models) {
                 if (resultClass != null) {
                     if (resultClass.equals(String.class)) {
                         object = (T) row.stringValue();
+                    } else if (resultClass.getPackage().getName().equals("java.lang") ||
```
mmichalek, line 664: so this is for a query that's like "select max(start_date) from trn_trans; kind of query?  What happens is the data in the row is not the right type.  Maybe we should do a type check there and throw an exception if it's not compatible.

liambell-le (approved):

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
chenson42: that is a good idea

PR #1064, opened 2021-01-12
Fixed Appriss not prompting to manually re enter

blubaugha:
### Issues Fixed
[PDPOS-4694](https://petcoalm.atlassian.net/browse/PDPOS-4694)

### Summary
This fixes an issue where the Appriss response JSON was not correctly parsed because this response uses different casing than for the properties than all other cases.

### Screenshots
appriss-re-enter-id

### DIFF
--- a/openpos-util/src/main/java/org/jumpmind/pos/util/DefaultObjectMapper.java
+++ b/openpos-util/src/main/java/org/jumpmind/pos/util/DefaultObjectMapper.java
@@ -4,6 +4,7 @@ import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.TimeZone;
 
+import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import org.joda.money.Money;
 import org.jumpmind.pos.util.web.MoneyDeserializer;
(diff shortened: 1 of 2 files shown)


openpos-util/src/main/java/org/jumpmind/pos/util/DefaultObjectMapper.java
```diff
@@ -27,6 +28,7 @@ public static ObjectMapper defaultObjectMapper() {
         mapper.setSerializationInclusion(Include.NON_NULL);
         mapper.registerModule(module);
         mapper.registerModule(new JavaTimeModule());
+        mapper.configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES, true);
```
jrmihalick, line 31: I don't think we want to do that here actually.  You can just set that attribute on the mapper instance you get back in the Appriss code.  Make sense?

jrmihalick (changes requested):

openpos-util/src/main/java/org/jumpmind/pos/util/DefaultObjectMapper.java
blubaugha: Ahhhh, yeah good point!

blubaugha:
Moving this change to verify analyze in commerce

PR #1065, opened 2021-01-13
Update item detail screen to support promo prices

eddielegg-le:
### Issues Fixed
https://petcoalm.atlassian.net/browse/PDPOS-3991

### Summary
Update Item Detail screen to include price point as well as promo price for all promos for a specified item. Separated promos into separate card on detail screen:

### Screenshots
image

image

No promotions:

image

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/item-detail/item-detail.component.html
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/item-detail/item-detail.component.html
@@ -31,7 +31,7 @@
                 <div class="promotion-item-name">
                     {{screen.itemValueDisplay.label}}
                 </div>
-                <div class="promotion-item-rewards">
+                <div class="price">
                     {{screen.itemValueDisplay.value}}
                 </div>
             </span>
@@ -43,12 +43,7 @@
                     <div class="promotion-item-name">
                         {{promo.promotionName}}
                     </div>
-                    <ul responsive-class class="promotion-item-rewards">
-                        <li *ngFor="let reward of promo.rewards">
-                            {{reward}}
-                        </li>
-                    </ul>
-                    <div class="promotion-item-name">
+                    <div class="price">
                         {{promo.promotionPrice}}
                     </div>
                 </span>
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/item-detail/item-detail.component.html
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/item-detail/item-detail.component.html
@@ -1,41 +1,59 @@
 <app-bacon-strip></app-bacon-strip>
-<section responsive-class class="item-card">
-    <h1 responsive-class>{{screen.itemName}}</h1>
-    <p responsive-class>{{screen.summary}}</p>
-    <app-carousel class="carousel" carouselSize="xl" [carouselItemClass]="'rounded-edge'">
-        <ng-template *ngFor="let image of screen.imageUrls" #carouselItem>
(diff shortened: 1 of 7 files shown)


stojankz (changes requested):
Would it be possible to make the promotions section span across the entire width if there is no other section to be shown on the screen? Also, looks like there might need to be some additional styling updates to make the prices line up with each other on the promotional lines. Could probably be achieved using grid area

eddielegg-le:
> Would it be possible to make the promotions section span across the entire width if there is no other section to be shown on the screen? Also, looks like there might need to be some additional styling updates to make the prices line up with each other on the promotional lines. Could probably be achieved using grid area.. or you could probably remove the reward descriptions? I don't think those are really needed any more?

It was requested to have it only be on the right side of the screen. Also, the Selling Price line will have fewer items than the promo lines (Selling Price will have title and amount, while promo will have icon, title, price point, and promo price)

stojankz:
> > Would it be possible to make the promotions section span across the entire width if there is no other section to be shown on the screen? Also, looks like there might need to be some additional styling updates to make the prices line up with each other on the promotional lines. Could probably be achieved using grid area.. or you could probably remove the reward descriptions? I don't think those are really needed any more?
> 
> It was requested to have it only be on the right side of the screen, and to include what was existing previously on promos plus the promo price. Also, the Selling Price line will have fewer items than the promo lines (Selling Price will have title and amount, while promo will have icon, title, price point, and promo price)

If it was requested to just be on the right that's fine. But I think you can still probably use grid area to make sure the columns match up with each other even when a row doesn't have some of the column details.

eddielegg-le:
> > > Would it be possible to make the promotions section span across the entire width if there is no other section to be shown on the screen? Also, looks like there might need to be some additional styling updates to make the prices line up with each other on the promotional lines. Could probably be achieved using grid area.. or you could probably remove the reward descriptions? I don't think those are really needed any more?
> > 
> > 
> > It was requested to have it only be on the right side of the screen, and to include what was existing previously on promos plus the promo price. Also, the Selling Price line will have fewer items than the promo lines (Selling Price will have title and amount, while promo will have icon, title, price point, and promo price)
> 
> If it was requested to just be on the right that's fine. But I think you can still probably use grid area to make sure the columns match up with each other even when a row doesn't have some of the column details.

They're using the grid. I changed the cell for the promo icon from "auto" to "1fr" so they will all be consistent:

image

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/item-detail/item-detail.component.html
```diff
@@ -1,41 +1,59 @@
...
+                          [instructionsSize]="'text-md'"></app-instructions>
+        <div>
+            <span class="promotion-items promotion-item-gray">
                 <div class="promotion-item-icon">
-                    <app-icon [iconName]="promo.icon" color="primary" class="promotion-icon" iconClass="md"></app-icon> 
+                    <app-icon iconClass="md"></app-icon>
                 </div>
                 <div class="promotion-item-name">
-                    {{promo.promotionName}}
+                    {{screen.itemValueDisplay.label}}
```
stojankz, line 32: Part of the issue is probably that this itemValueDisplay is separate from the list of other promotions. The data for the selling price should probably be passed in as part of the list of promotions

eddielegg-le: Right, that was intentional. It's not really a promo...it'll only have a title and a value, which worked well with the existing DisplayProperty class.

jrmihalick (changes requested):
See my comments on https://github.com/JumpMind/commerce/pull/2853.

Sorry I put my comment on the wrong PR.

jrmihalick (approved):

jrmihalick (approved):

PR #1066, opened 2021-01-13
Oracle creates an array of the size of the fetchSize.  It needs to be a reasonable number.

chenson42:
### Summary
Oracle creates an array of the size of the fetchSize.  It needs to be a reasonable number.


dankaste:
should this be configurable? Would it ever need to be > 10000?

chenson42:
> should this be configurable? Would it ever need to be > 10000?

I don't think it needs to be.  a max fetch size of 10000 sounds reasonable.

dankaste (approved):

PR #1072, opened 2021-01-17
Require that flow extensions have the suffix of *-flow-ext.yml

chenson42:
### Summary
In order to account for classpath differences from platform to platform and environment to environment require that flow extensions have a different suffix so that we can explicitly load them last and provide warnings to developers when they are misconfigured.


mmichalek (approved):

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/config/YamlConfigProvider.java
```diff
@@ -47,14 +47,15 @@ public void load(String appId, String path) {
         try {
             ResourcePatternResolver resolver = new PathMatchingResourcePatternResolver(Thread.currentThread().getContextClassLoader());
 
-            Resource[] resources = resolver.getResources("classpath*:/" + path + "/*-flow.yml");
+            List<Resource> resources = new ArrayList<>(Arrays.asList(resolver.getResources("classpath*:/" + path + "/*-flow-ext.yml")));
+            resources.addAll(Arrays.asList(resolver.getResources("classpath*:/" + path + "/*-flow.yml")));
 
             List<YamlFlowConfig> yamlFlowConfigs = new ArrayList<>();
 
             // Loop these backwards since most specific will be first in the class path an we want those
```
jrmihalick, line 55: This comment doesn't seem valid any longer?  The reason for the fix is because depending on the classpath is not dependable.  :-)

jrmihalick (changes requested):

PR #1078, opened 2021-01-18
Feature/training mode watermark

jchirico8:
### Summary
Adding a watermark to the page to denote that you are in training mode. This is toggled by a call from TrainingModeState through the UIManager to send a message to the angular side. I added the watermark to the dynamic-screen component since that is the parent for all screens and we want this to persist as long as you are still in training mode. I'm tagging @blubaugha to double check my angular changes and @JoeBrownlee since he's working on other training related changes.

Dependent on JumpMind/commerce#2876

### Screenshots
image
image
image


openpos-flow/src/main/java/org/jumpmind/pos/core/ui/HideWatermark.java
```diff
@@ -0,0 +1,11 @@
+package org.jumpmind.pos.core.ui;
+
+import org.jumpmind.pos.core.model.MessageType;
+import org.jumpmind.pos.util.model.Message;
+
+public class HideWatermark extends Message {
```
jrmihalick, line 6: Is there a reason we have two message classes, one to hide the watermark and one to show the watermark, instead of just one message with a boolean?

jchirico8: It was modeled after Toast, which has a CloseToast message as well. Mark and I agreed that this is excessive for the watermark use case, but it's consistent with how we handle toast so I left it.

jchirico8:
Closing since this was merged separately into a dedicated training mode branch

PR #1081, opened 2021-01-20
Initial Implementation of Status Bar Details

kdelorey:
# Summary
The Status bar can now be interacted with on **any** screen. A simple click will bring up a modal dialog. The intent is that the dialog will one day show details about status items and the server. Right now it does not do that but what it does do is provide a way to select a Printer and Payment device that shall be used. The status bar will display the currently selected printer/payment device too.

Note that this is just an initial implementation to support printer/payment selection. The status details still need to be implemented and currently lack full support for displaying on this dialog. There is also some styling considerations that need to be reviewed later too.

### Overlay Container
The CDK overlay container has been overridden as to not affect the entire screen but rather a selected portion of the screen.

### Status Move
The Status implementation was previously within the core folder but loaded through `SharedModule`. I have moved all of the status bar relevant items to the `shared` folder where it should live with its parent module.

### New `localize` Pipe
A new Angular pipe has been added that allows you too fulfill a i18n key from the client. It accepts a `base:key` format and uses a new `getString()` function on the `LocaleService` to invoke the `/i18n/value` endpoint on the server.

```html
<app-some-component [title]="'page:some.key.example' | localize">
  ....
```

This pipe will automatically respond to changes in locale too. Finally the pipe all supports i18n arguments too.

### Screenshots
Example


kdelorey:
I added pretty much everyone I could think of as a reviewer just so everyone could see some of these important changes. Its probably not required that everyone actually review it, but you may want to be familiar with the changes.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/messages/message-types.ts
```diff
@@ -20,4 +20,5 @@ export class MessageTypes {
     static AUDIO_CONFIG = 'AudioConfig';
     static AUDIO_PRELOAD = 'AudioPreload';
     static LOADING = 'Loading';
+    static PERIPHERAL_DEVICE_SELECTION = "PeripheralDeviceSelection";
```
blubaugha, line 23: Sorry to nitpick 😄 , but the double-quotes should be replaced with single quotes

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/messages/peripheral-device-selection.ts
```diff
@@ -0,0 +1,20 @@
+import { MessageTypes } from "./message-types";
+import { OpenposMessage } from "./message";
```
blubaugha, line 2: Replace these double-quotes with single quotes too

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/locale.service.ts
```diff
@@ -1,28 +1,38 @@
...
-    constructor(public sessionService: SessionService) {
+    private locale$ = new BehaviorSubject<string>(DEFAULT_LOCALE);
+    private displayLocale$ = new BehaviorSubject<string>(DEFAULT_LOCALE);
+
+    constructor(
+        public sessionService: SessionService,
+        private http: HttpClient,
+        private personalization: PersonalizationService
+    ) {
         this.sessionService.getMessages('LocaleChanged').subscribe(message => this.handleLocaleChanged(message));
```
blubaugha, line 27: If the server is set to a locale other than **en-US**, will this receive a `LocaleChanged` message on startup so the initial locale matches the server?

blubaugha (changes requested):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/messages/message-types.ts
kdelorey: Oh yeah good catch, I try to single quote in typescript but sometimes my hands get confused when switching between languages.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/locale.service.ts
kdelorey: It should. From what I observed a `'LocaleChanged '` event is observed on connect. This will work if the locale is changed on the fly too. I haven't change the behavior of this from what it was before.

blubaugha: 👍

maxwellpettit (approved):

JJStarrett (approved):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/messages/peripheral-device-selection.ts
kdelorey: Fixed.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/messages/message-types.ts
kdelorey: Fixed.

blubaugha (approved):

blubaugha:
By the way, nice job on this!

stevencarley (approved):

PR #1089, opened 2021-01-25
Feature/client executable action

eddielegg-le:
### Issues Fixed
https://petcoalm.atlassian.net/browse/PDPOS-2542

### Summary
Create a workflow that will allow the client to execute commands (either shell command or execute file) through the client wrapper (Electron, for example). 

Notes:
The `IActionItem` sent to the client will require the `action` field to be "ClientExecutable", and the `defaultPayload` value should by the name of the client execute config yaml. For example, the IActionItem may be:
```
{
    "action" : "ClientExecutable",                <----
    "title" : "Timeclock",
    "defaultPayload" : "timeclock",        <----
    "icon" : "Calendar",
    "enabled" : true,
    "sensitive" : false,
    "queueIfBlocked" : false,
  }
```
and the configuration may be:
```
  clientExecutable:
    timeclock:  <-----
      data: 'do stuff'
      executableType: Command
```

The  configuration can be deserialized into a `ClientExecutableConfigMessage` and will need the following values in order to correctly execute commands:
- `executableType`: Enum values of `Command` or `File` to denote which to execute
- `data`: The command to execute or the filepath (depending on the executable type)
- `args`: Optional array of string args to pass to the command

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/actions/action.service.ts
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/actions/action.service.ts
@@ -25,10 +25,10 @@ export class ActionService implements OnDestroy {
         private messageProvider: MessageProvider) {
         console.log("Creating new Action Service")
         this.subscriptions.add(messageProvider.getScopedMessages$().subscribe(message => {
-            if(message.willUnblock === false) {
+            if (message.willUnblock === false) {
                 console.log('creating a screen that is disabled');
                 this.blockActions = true;
-            } else if( message.willUnblock) {
+            } else if (message.willUnblock) {
                 console.log('unblocking actions because message:', message);
                 this.unblock();
             }
@@ -38,21 +38,20 @@ export class ActionService implements OnDestroy {
                 console.log('unblocking action because toast:', message);
                 this.unblock();
             }
+            if (message.type === MessageTypes.CLIENT_EXECUTE) {
+                console.log('unblocking action because client execute:', message);
+                this.unblock();
+            }
         }));
-    }
-
-    private unblock() {
-        this.blockActions = false;
-        const queued = this.actionQueue.pop();
-        if (queued) {
-            console.log('Dequeued an action to send');
-            this.doAction(queued.item, queued.payload);
-        }
     }
 
     async doAction(actionItem: IActionItem, payload?: any) {
         const sendAction = await this.canPerformAction(actionItem);
         if (sendAction) {
+            if (payload && actionItem.defaultPayload) {
+                console.warn("Default action payload overridden for action " + actionItem.action);
+            }
+            payload = payload || actionItem.defaultPayload;
             if (!actionItem.doNotBlockForResponse) {
                 this.blockActions = true;
             }
@@ -82,16 +81,9 @@ export class ActionService implements OnDestroy {
         }
     }
 
-    private doUrlAction(urlItem: IUrlMenuItem) {
-        // check to see if we are an IURLMenuItem
-        console.info(`About to open: ${urlItem.url} in target mode: ${urlItem.targetMode}, with options: ${urlItem.options}`);
-        window.open(urlItem.url, urlItem.targetMode, urlItem.options);
-    }
-
     public actionBlocked(): boolean {
         return this.blockActions;
     }
-
 
     public registerActionPayload(actionName: string, actionValue: () => void) {
         this.actionPayloads.set(actionName, actionValue);
@@ -127,6 +119,27 @@ export class ActionService implements OnDestroy {
         }
 
         return this.actionDisablers.get(action).value;
+    }
+
+    ngOnDestroy(): void {
+        if (this.subscriptions) {
+            this.subscriptions.unsubscribe();
+        }
+    }
+
+    private unblock() {
+        this.blockActions = false;
+        const queued = this.actionQueue.pop();
+        if (queued) {
+            console.log('Dequeued an action to send');
+            this.doAction(queued.item, queued.payload);
+        }
+    }
+
+    private doUrlAction(urlItem: IUrlMenuItem) {
+        // check to see if we are an IURLMenuItem
+        console.info(`About to open: ${urlItem.url} in target mode: ${urlItem.targetMode}, with options: ${urlItem.options}`);
+        window.open(urlItem.url, urlItem.targetMode, urlItem.options);
     }
 
     private queueLoading() {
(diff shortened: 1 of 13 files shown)


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/actions/action.service.ts
```diff
@@ -38,21 +38,20 @@ export class ActionService implements OnDestroy {
...
                 this.unblock();
             }
+            if (message.type === MessageTypes.CLIENT_EXECUTE) {
+                console.log('unblocking action because client execute:', message);
+                this.unblock();
+            }
         }));
     }
 
-    private unblock() {
```
eddielegg-le, line 44: IDE did a lot of auto formatting to conform to the linter

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/actions/action.service.ts
```diff
@@ -38,21 +38,20 @@ export class ActionService implements OnDestroy {
                 console.log('unblocking action because toast:', message);
                 this.unblock();
             }
+            if (message.type === MessageTypes.CLIENT_EXECUTE) {
```
eddielegg-le, line 41: Added this to avoid blocking another client execute action

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/toast.service.ts
```diff
@@ -6,33 +6,32 @@ import {ToastComponent} from "../../shared/components/toast/toast.component";
...
 export class ToastService {
-    persistedToasts = new Map<String,ActiveToast<any>>();
-    constructor( private sessionService: SessionService, private toastrService: ToastrService ) {
+    persistedToasts = new Map<String, ActiveToast<any>>();
+
+    constructor(private sessionService: SessionService, private toastrService: ToastrService) {
         sessionService.getMessages('Toast').subscribe(m => this.showMessage(m));
         sessionService.getMessages('CloseToast').subscribe(m => this.closeMessage(m));
-        sessionService.getMessages('Connected').subscribe(m => this.toastrService.clear());
-        window['toastService'] = this;
```
eddielegg-le, line 16: This shouldn't have made it into the PR, so removing it

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/actions/action.service.ts
jrmihalick: @dankaste: Should we handle this way or should we just set the default value of `willUnblock` for the ClientExecuteMessage on the server side to `true`?  I am confused on the the purpose of `willUnblock`.  We need to document it on the Message class.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/core.module.ts
```diff
@@ -165,6 +166,7 @@ export class CoreModule {
                 logger: ConsoleIntercepter,
                 toastService: ToastService,
                 uiDataService: UIDataMessageService,
+                clientExecuteService: ClientExecuteService,
```
jrmihalick, line 169: After looking at this in code, do you think ClientExecutableService would be a better name?  ClientExecuteService makes me think of executing the client.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/interfaces/client-execute-config.interface.ts
```diff
@@ -0,0 +1,5 @@
+export interface ClientExecuteConfig {
```
jrmihalick, line 1: Likewise, ClientExecutableConfig ?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/client-execute.service.ts
```diff
@@ -0,0 +1,28 @@
...
+    private execute(message: any) {
+        const config = message as ClientExecuteConfig;
+        console.log("Executing command...")
+        const args = _.concat([config.data], config.args);
+        if (this.electronService.isElectronApp) {
+            this.electronService.ipcRenderer.invoke(config.executableType, args);
+        } else {
+            console.warn('Cannot execute command ' + config.data);
+        }
+        this.sessionService.cancelLoading();
```
jrmihalick, line 26: If we set `doNotBlockForResponse=true` on the server side ActionItem, is this still needed?

openpos-flow/src/main/java/org/jumpmind/pos/core/ui/ActionItem.java
```diff
@@ -152,13 +153,20 @@ public ActionItem title(String title) {
...
         return this;
     }
-    
-    
+
+    public String getDefaultPayload() {
+        return defaultPayload;
+    }
+
+    public void setDefaultPayload(String defaultPayload) {
```
jrmihalick, line 161: Could you add some Javadoc to describe what the defaultPayload is used for?  Thanks, it may help future time travelers.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/actions/action.service.ts
eddielegg-le: Yeah I like the `willUnblock` route better...I was thinking this needed passed down on the ActionItem and would need special logic for that, but it makes sense to default it on the ClientExecutableMessage instead.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/core.module.ts
eddielegg-le: I actually originally had it ad ClientExecutableXXX but changed it 😆  I can change it back though

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/client-execute.service.ts
eddielegg-le: Unfortunately not unless we set `doNotBlockForResponse` = true on the action item, which I'd prefer not to do.

openpos-flow/src/main/java/org/jumpmind/pos/core/ui/ActionItem.java
eddielegg-le: Indeederino

jrmihalick (approved):

PR #1094, opened 2021-01-27
Initial Release of Enhanced Training Mode Infrastructure

JoeBrownlee:
### Issues Fixed
This feature addresses the following Petco JIRA items in whole or in part: #PDPOS-2759, #PDPOS-2823, #PDPOS-2824, #PDPOS-4117, #PDPOS-4241, #PDPOS-4243.

### Summary
This adds a greatly enhanced version of Training Mode to the OpenPOS Framework and Commerce. Users can train on most every function of the POS using real or "shadowed" data and services. This is highly configurable.

The basis of this change is a new concept of a **Device Mode**. The Device Mode for the POS is normally `default`. When the POS is in Training Mode, it is `training`. Other modes could be added in the future.

When in Training Mode, the POS can use **Shadow Tables** and **Shadow Endpoints**. Shadow tables are parallel versions of Commerce tables that are separate from the main production database tables. They have the same name and structure as the "normal" tables, except that they are prefixed with a special shadow prefix, `TNG` by default. The prefix is configurable. The shadow tables are used while the POS is in Training Mode.

Likewise, shadow endpoints can be created to "override" normal POS behavior while in Training Mode. A shadow endpoint will have the same service path as the default endpoint, but its implementation will be `training`.

The `application.yml` file in the commerce-server project controls the configuration of shadow tables. Its `deviceMode` tag must match the Training Mode value `training`. It specifies the shadow table prefix, whether queries should be validated at startup, and tables to be included or excluded from shadowing. `*` can be used as a wildcard in table names. For example:
```
shadowTables:
    deviceMode: training
    tablePrefix: tng
    validateQueries: true
    includes:
      - sls_*
      - cust_*
      - ops_*
    excludes:
      - sls_rtn_policy
      - sls_tender
      - cust_phone_blacklist
  ```

For the POS to adjust SQL to use the shadow tables, the `*-query.yml` files need to have their SQL written to use the model class name for a given entity rather than a hard-coded table name. For example, a query like `select * from SLS_TRANS` would be written instead as `select * from TransModel`. As part of this work, all existing SQL in the query files has been changed to conform to this. Note _one important exception_: if a query has a join or sub select that uses a table outside its own module, those table names must be left as the actual table name rather than the model class name. For example, in `sls-query.yml`, there is SQL that selects data from the `USR_USER` table. That reference must stay as the actual table name.

The GUI has been changed as well. In previous versions of Commerce, clicking the **Training Mode** button on the Home screen would allow the operator to perform one transaction and then they would exit Training Mode. Now, when the operator clicks the **Training Mode** button, the POS will be in Training Mode until the operator explicitly exits by returning to the Home screen and clicking the **Training Mode** button again. Entering Training Mode requires a manager's authorization. While in Training Mode, the color scheme changes to a grey palette and a watermark appears at the top of each screen to indicate the POS is in Training Mode.

This is the initial version of this feature. Future PRs will add additional functionality.

### Database Changes
There are no database changes per se, other than the fact that shadow tables will be created automatically at startup as specified in the `application.yml` file. Existing tables are not changed.

### Screenshots
Screen Shot 2021-01-27 at 12 06 55 PM
Screen Shot 2021-01-27 at 12 06 33 PM
Screen Shot 2021-01-27 at 12 07 43 PM

### CODE REVIEW

File: openpos-devices/src/main/java/org/jumpmind/pos/devices/model/DeviceModel.java [hunk 1]
```diff
@@ -51,6 +53,30 @@
 
     @ColumnDef(size = "255", description = "A user defined name for the Device")
     private String description;
+    @ToString.Include
+    @Builder.Default
+    @ColumnDef(size = "100", description = "The current mode of the Device")
+    private String deviceMode = DEVICE_MODE_DEFAULT;
```
jrmihalick, line 59: I'm wondering if we should make this a TypeCode.  If the type of this were DeviceModeTypeCode instead of String, would that make life harder?

jrmihalick: Might cause issues with your changes in StateManagerContainer.

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSessionFactory.java
```diff
@@ -55,67 +76,95 @@ public void init(
...
                 log.info(String.format("Loading %s...", url.toString()));
                 InputStream queryYamlStream = url.openStream();
                 QueryTemplates queryTemplates = new Yaml(new Constructor(QueryTemplates.class)).load(queryYamlStream);
-                return queryTemplates;
-            } else {
-                log.debug("Could not locate " + tablePrefix + "-query.yml on the classpath.");
-                return new QueryTemplates();
+                if (queryTemplates != null) {
+                    templates.addQueries("default", queryTemplates.getQueries());
+                    templates.addQueries("training", queryTemplates.getQueries());
```
jrmihalick, line 160: Would be nice if this could use the same constant values in DeviceModel, but DeviceModel is in another module.

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSessionFactory.java
```diff
@@ -217,20 +340,23 @@ protected void warnOrphanedTagColumns(List<TagModel> tags, Table table) {
...
+    protected Column generateTagColumn(TagModel tag, Table table, boolean modifyPk) {
+        return setColumnInfo(new Column(), table, tag, modifyPk);
     }
 
-    protected Column setColumnInfo(Column column, TagModel tag, boolean includeInPk) {
+    protected Column setColumnInfo(Column column, Table table, TagModel tag, boolean includeInPk) {
         column.setName(getColumnName(tag));
         column.setPrimaryKey(includeInPk);
+        if (includeInPk) {
+            // column.setPrimaryKeySequence(table.getPrimaryKeyColumnCount() + 1);
```
jrmihalick, line 351: Should this if stmt be removed?

openpos-persist/src/main/java/org/jumpmind/pos/persist/impl/DatabaseSchema.java
```diff
@@ -444,14 +522,15 @@ private static Column createColumn(Field field, IDatabasePlatform platform, Mode
 
             if (metaData.isPrimaryKey(field)) {
                 dbCol.setRequired(true);
+                // dbCol.setPrimaryKeySequence(new ArrayList<>(metaData.getPrimaryKeyFieldNames()).indexOf(dbCol.getName()));
```
jrmihalick, line 525: Remove?

openpos-persist/src/main/java/org/jumpmind/pos/persist/impl/ModelClassMetaData.java
```diff
@@ -50,16 +57,39 @@ public void init() {
...
 
     private void initPrimaryKeyColumns() {
+        int primaryKeySequence = 0;
         for (String primaryKeyFieldName : primaryKeyFieldNames) {
             FieldMetaData fieldMetaData = entityIdFieldMetaDatas.get(primaryKeyFieldName);
             if (fieldMetaData == null) {
                 throw new PersistException("There is no field meta data for field name '" + primaryKeyFieldName + "' for model " + getClazz().getSimpleName());
             }
 
+            // fieldMetaData.getColumn().setPrimaryKeySequence(primaryKeySequence++);
```
jrmihalick, line 67: Remove?

openpos-service/src/main/java/org/jumpmind/pos/service/AbstractRDBMSModule.java
```diff
@@ -78,14 +95,23 @@
     @Value("${openpos.general.failStartupOnModuleLoadFailure:false}")
     boolean failStartupOnModuleLoadFailure;
 
+    //  Per CEH, the below is needed for AEO only.
```
jrmihalick, line 98: ??

openpos-persist/src/test/java/org/jumpmind/pos/persist/cars/TestPersistCarsConfig.java
```diff
@@ -48,16 +45,61 @@ public DBSessionFactory sessionFactory() {
             
             TagHelper tagHelper = new TagHelper();
             tagHelper.setTagConfig(tagConfig);
-            
+
+            AugmenterConfigs augmenterConfigs = new AugmenterConfigs();
```
jrmihalick, line 49: Are there any unit tests which test a simple shadow table configuration?

jrmihalick (changes requested):
Some of my comments are just questions, others are minor changes.  This is a ton of work.  Overall a really nice PR.

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSessionFactory.java
JoeBrownlee: This came over from Commerce master but we don't have the most recent version Column on 1.9. So we can get rid of it for now, but it will come back.

openpos-persist/src/main/java/org/jumpmind/pos/persist/impl/DatabaseSchema.java
JoeBrownlee: This came over from Commerce master but we don't have the most recent version Column on 1.9. So we can get rid of it for now, but it will come back.

openpos-persist/src/main/java/org/jumpmind/pos/persist/impl/ModelClassMetaData.java
JoeBrownlee: This came over from Commerce master but we don't have the most recent version Column on 1.9. So we can get rid of it for now, but it will come back.

openpos-service/src/main/java/org/jumpmind/pos/service/AbstractRDBMSModule.java
JoeBrownlee: Chris said the autowire is not needed other than in AEO. Commerce/Petco definitely don't run with that line in place. Don't ask me to explain that. I just know it doesn't work with that enabled.

openpos-persist/src/test/java/org/jumpmind/pos/persist/cars/TestPersistCarsConfig.java
JoeBrownlee: As of this moment, there are no tests, but I haven't gotten to the Cucumber task yet.

openpos-devices/src/main/java/org/jumpmind/pos/devices/model/DeviceModel.java
JoeBrownlee: In general, I'd like to see that constant be in a place that has more visibility. Too many places can't see it in DeviceModel, therefore you end up hard-coding it which defeats the purpose. Not sure the right place, but I could look into that -- unless someone wants to give me the right answer!

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSessionFactory.java
JoeBrownlee: Exactly, I agree. Open to suggestions...

openpos-devices/src/main/java/org/jumpmind/pos/devices/model/DeviceModel.java
jrmihalick: After looking at the rest of your PR, probably shouldn't mess with this.  You can ignore my comment here

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSessionFactory.java
jrmihalick: Disregard as per my other comment above

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSessionFactory.java
jrmihalick: Ok, I see, don't worry about it

openpos-persist/src/main/java/org/jumpmind/pos/persist/impl/DatabaseSchema.java
jrmihalick: Same as other, don't worry about it

openpos-persist/src/main/java/org/jumpmind/pos/persist/impl/ModelClassMetaData.java
jrmihalick: Same

openpos-persist/src/test/java/org/jumpmind/pos/persist/cars/TestPersistCarsConfig.java
jrmihalick: ok, once we get this PR put to bed, let's try to get a unit test or two put in before Cucumber tests.  We might have to drop kick cucumber tests until next release.

jrmihalick (approved):

PR #1109, opened 2021-02-05
Prevent confusing subflow reference to another standalone state

mmichalek:
### Summary
Prevent confusing subflow reference to another standalone state


blubaugha (approved):

jrmihalick (approved):

jrmihalick:
👍

PR #1116, opened 2021-02-09
Make displaying peripheral selection in status bar configurable

levshunovm:
### Summary
Make displaying peripheral selection in status bar configurable.

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/status/status-bar/status-bar.component.ts
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/status/status-bar/status-bar.component.ts
@@ -1,6 +1,6 @@
 import { Component, ElementRef, HostListener, Renderer2 } from '@angular/core';
 import { combineLatest, Observable, of } from 'rxjs';
-import { map, switchMap, tap } from 'rxjs/operators';
+import { map, publishBehavior, refCount, switchMap, tap } from 'rxjs/operators';
 import { ConfigChangedMessage } from '../../../core/messages/config-changed-message';
 import { Status } from '../../../core/messages/status.enum';
 import { StatusMessage } from '../status.message';
@@ -23,8 +23,6 @@ export class StatusBarComponent {
   systemInfo$: Observable<SystemInfo>;
 
   line1$: Observable<string>;
-
-  selectable = false;
 
   constructor(
     private statusService: StatusService,
@@ -61,25 +59,23 @@ export class StatusBarComponent {
               )
             }
           )
-        ))
+        )),
+        publishBehavior(new Array<CategoryWithName>()),
+        refCount()
       )
     ).pipe(
       map(results => {
         let l = results[0].line1;
         const i = results[1];
 
-        this.selectable = false;
         i.forEach((value) => {
-          if (value.selectable) {
-            this.selectable = true;
-            let dn = "Not Selected";
+          let dn = "Not Selected";
 
-            if (value.selectedDevice && value.selectedDevice.displayName) {
-              dn = value.selectedDevice.displayName;
-            }
-
-            l += " | " + value.displayName + " " + dn;
+          if (value.selectedDevice && value.selectedDevice.displayName) {
+            dn = value.selectedDevice.displayName;
           }
+
+          l += " | " + value.displayName + " " + dn;
         });
 
         return l;
@@ -89,9 +85,7 @@ export class StatusBarComponent {
 
   @HostListener('click')
   onClick() {
-    if (this.selectable) {
-      this.statusService.openDetails();
-    }
+    this.statusService.openDetails();
   }
 }
 
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/status/status-bar/status-bar.component.ts
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/status/status-bar/status-bar.component.ts
@@ -24,6 +24,8 @@ export class StatusBarComponent {
 
   line1$: Observable<string>;
 
+  selectable = false;
+
   constructor(
     private statusService: StatusService,
     peripheralSelectionService: PeripheralSelectionService,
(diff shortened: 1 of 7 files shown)


kdelorey:
@chenson42 did we actually want this configurable? I assumed this would lead into the status details page always, but I guess too be fair, that's not done yet and I suspect its not on the radar.

kdelorey:
The impetus for this PR is actually all caused by a bug in the backed server. The only thing this configuration should do is prevent the status details dialog from appearing.

The backend server shouldn't be sending peripheral messages to the client when not in mobile mode. This will give the possibility in the future to have selectable peripherals for non-mobile devices in the future; but I realize that this not a goal, I just think it is something we can easily have just in case.

Reach out to me and I can point you to what needs to happen in the backend.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/status/status-bar/status-bar.component.ts
```diff
@@ -83,7 +89,9 @@ export class StatusBarComponent {
 
   @HostListener('click')
   onClick() {
-    this.statusService.openDetails();
+    if (this.selectable) {
```
kdelorey, line 92: You have whether or not the status details dialog is shown based on if peripheral selection exists. This should probably not be the case since the status details is intended to be something greater than just peripheral selection. You may want a separate configuration that lives with the status stuff for now.

Also "selectable" can be entirely inferred. If you don't receive any messages then you can assume `false`. You can assume `|= true` when you receive a message with a selectable array size > 0.

kdelorey (changes requested):

levshunovm:
@kdelorey I tried to just disable these messages in the backend, but it leads to the problems with combineLatest on lines 42-77 in status-bar.component.ts. It isn't working when one of the arguments is empty, so it disables the whole line in the status including store and device number. I tried a lot of different things, but still can't find a way to make it works. So, could you please advice how to fix it?

kdelorey:
> @kdelorey I tried to just disable these messages in the backend, but it leads to the problems with combineLatest on lines 42-77 in status-bar.component.ts. It isn't working when one of the arguments is empty, so it disables the whole line in the status including store and device number. I tried a lot of different things, but still can't find a way to make it works. So, could you please advice how to fix it?

Something like this should do it. This is off the top of my head, so you may need to massage it a little, but `publishBehavior()` will always make sure there's a value there when its first subscribed to.
```ts
peripheralSelectionService.peripheralCategories$.pipe(
    switchMap(categories => combineLatest( ... )),
    publishBehavior(new Array<CategoryWithName>()),
    refCount()
)
```

levshunovm:
> > @kdelorey I tried to just disable these messages in the backend, but it leads to the problems with combineLatest on lines 42-77 in status-bar.component.ts. It isn't working when one of the arguments is empty, so it disables the whole line in the status including store and device number. I tried a lot of different things, but still can't find a way to make it works. So, could you please advice how to fix it?
> 
> Something like this should do it. This is off the top of my head, so you may need to massage it a little, but `publishBehavior()` will always make sure there's a value there when its first subscribed to.
> 
> ```ts
> peripheralSelectionService.peripheralCategories$.pipe(
>     switchMap(categories => combineLatest( ... )),
>     publishBehavior(new Array<CategoryWithName>()),
>     refCount()
> )
> ```

Thanks! It works. I also tried to fix all your other comments in both repos. Could you please re-review?

kdelorey (changes requested):
There's a pretty big issue with mixing asynchronous code with synchronous code that needs to be looked at again. Other than that I think everything seems good.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/status/status-details/status-details.service.ts
```diff
@@ -0,0 +1,17 @@
...
+export class StatusDetailsService {
+  constructor(
+    public peripheralSelection: PeripheralSelectionService
+  ) {}
+
+  isDetailsEmpty() {
+    let count = 0;
+    this.peripheralSelection.peripheralCategories$.subscribe(value => count += value.length);
+    return count == 0;
+  }
```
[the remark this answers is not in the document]
levshunovm: It seems that when `peripheralCategories$` is empty, the pipe's body is ignored and function just returns observable with false, which is totally fine in our case, but I had to invert the function and rename it. 
Also, I haven't found a way to work with `Observable<boolean>` without `subscribe`, so not sure if I fixed an issue with mixing asynchronous code with synchronous code that you've mentioned above.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/status/status.service.ts
```diff
@@ -38,6 +40,12 @@ export class StatusService {
             return;
         }
 
+        let isDetailsEmpty: boolean;
+        this.statusDetailsService.isDetailsNotEmpty().subscribe(value => isDetailsEmpty = !value);
+        if (isDetailsEmpty) {
+            return;
+        }
```
kdelorey, line 47: So you moved the mixing async with sync code problem here. You'll need to update this whole function to understand that you're doing async stuff here.

Something like this should take care of it; might not be perfect though.

```ts
public openDetails() {
    if (this.detailsDialog) {
        return;
    }

    this.statusDetailsService.isDetailsNotEmpty().pipe(
        take(1),
        filter(isDetailsNotEmpty => !isDetailsNotEmpty),
        map(() => {
            this.detailsDialog = this.dialog.open(StatusDetailsComponent, { minWidth: '75%' })
            return this.detailsDialog;
        }),
        mergeMap(dialog => dialog.afterClosed())
    ).subscribe({
        // don't need to worry about the subscription because the
        // observable will be automatically completed by the
        // source
        complete: () => {
            this.detailsDialog = undefined;
        }
    })
}
```

kdelorey (changes requested):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/status/status-details/status-details.service.ts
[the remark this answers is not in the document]
kdelorey: Ah, if you didn't want to change the name you could have just flipped the `boolean` expression to `v.length <= 0`. That was my bad. But regardless this should be stable now.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/status/status.service.ts
levshunovm: Thanks a lot for your help! I think it should be good now.

kdelorey (approved):

PR #1117, opened 2021-02-10
PCOM-1550 User is able to send mail to incorrect email format address

apomanytskyi:
Changed email validation rules


stevencarley:
Is there are reason this is needed over the the angular email validation? The samples I see on the ticket abc@xyz abc@xyz.xyz are both valid email address.

PR #1118, opened 2021-02-10
Style the status dialog similar to other dialogs (like the …

apomanytskyi:
Style the status dialog similar to other dialogs (like the options dialog)

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/status/status-details/status-details.component.html
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/status/status-details/status-details.component.html
@@ -1,14 +1,16 @@
 <div class="status-outer" responsive-class>
     <mat-card-title gdAreas="header backbutton" gdColumns="1fr auto" gdAlignColumns="center center">
         <div class="header" gdArea="header" gdColumns="auto 1fr" responsive-class>
-            <span>Status</span>
+            <span>Device Pairing</span>
         </div>
     </mat-card-title>
     <div class="status-body">
         <app-status-details-item
                 *ngFor="let s of peripheralSelection.peripheralCategories$ | async"
-                [title]="(s.localizationKey | localize) +  ' - ' + (s.selectedDevice?.displayName || 'Not Selected')"
+                [title]="s.selectedDevice?.displayName || (s.localizationNoCategorySelectedKey | localize)"
                 status="online"
+                [icon]="s.icon"
+                [titleTransparency]="s.selectedDevice?.displayName"
                 (actionExecuted)="onChangeSelectedPeripheral(s)">
         </app-status-details-item>
     </div>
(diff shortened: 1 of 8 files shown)


stevencarley:
Could you add screenshots?

apomanytskyi:
Screen Shot 2021-02-10 at 4 15 44 PM
Screen Shot 2021-02-10 at 4 16 15 PM

mmichalek:
Guys originally I had pictured the status screen as a cool "About" screen that also had a lot of useful status information. It seems like the screen pictured here is more of a device select screen, and not the status/about screen.  What do you think?
image

chenson42:
Here is another that Mark made with the devices.   If for right now we could do the device pairing portion of the screen that would be great.  Translucency is optional for this go round.   Basically do what you have only change Status to Device Pairing and introduce icons and button verbage.

image

apomanytskyi:
@chenson42 @mmichalek 
Ok, thanks.
Tell me if there is already a prototype of this page somewhere and can I see it? I mean may be html markup with styles...?
Or does this page only exist as a picture so far?

chenson42:
> @chenson42 @mmichalek
> Ok, thanks.
> Tell me if there is already a prototype of this page somewhere and can I see it? I mean may be html markup with styles...?
> Or does this page only exist as a picture so far?

There is not an html wireframe.   Like I mentioned, we can shortcut the design and use pretty much what you have and just change terminology and add icons for now.

kdelorey:
> @apomanytskyi Here is another screenshot that Mark made with the devices. If for right now we could do the device pairing portion of the screen that would be great. Translucency is optional for this go round. Basically what you have works with changing Status to Device Pairing and introducing icons and changing the button verbiage.
> 
> image

The backend was currently setup to pair directly with a status line item, so it won't have things like "icon" in the message. It could still be inferred from the pairing field I have, or just add a Icon field to the message.

apomanytskyi:
Depends on JMC PR #3068
For second part of screen I mock some data for pop up (that's why in toolbar I still have "Printer Not Selected")
Screen Shot 2021-02-11 at 6 27 20 PM

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/status/status-details/status-details.component.html
```diff
@@ -1,14 +1,16 @@
 <div class="status-outer" responsive-class>
     <mat-card-title gdAreas="header backbutton" gdColumns="1fr auto" gdAlignColumns="center center">
         <div class="header" gdArea="header" gdColumns="auto 1fr" responsive-class>
-            <span>Status</span>
+            <span>Device Pairing</span>
```
chenson42, line 4: This should be internationalized.   @dankaste @kdelorey How would we do this?

kdelorey: I added a `localize` pipe that does this for you given that there's a key on the server to match. Something along the lines of:

(Note I made this key up as an example)
```suggestion
            <span>{{status:details.dialog.title.deviceparing | localize}}</span>
```

kdelorey: More details about the pipe can be found in #1081

chenson42 (approved):

PR #1120, opened 2021-02-11
Move Loyalty Button

r-cochran:
### Summary
Moved the Customer Loyalty button from the sale total panel to a sub header section.

### Screenshots
Customer unlinked
image

Customer linked
image

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/icon.service.ts
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/icon.service.ts
@@ -21,6 +21,7 @@ export class IconService {
         private iconRegistry: MatIconRegistry) {
 
         this.icons.set('Account', ['account_balance.svg', this.defaultIconProvider]);
+        this.icons.set('AccountCircle', ['account_circle.svg', this.defaultIconProvider]);
         this.icons.set('Add', ['add_circle.svg', this.defaultIconProvider]);
         this.icons.set('AddCustomer', ['person_add.svg', this.defaultIconProvider]);
         this.icons.set('AddEmployee', ['person_add.svg', this.defaultIconProvider]);
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/icon.service.ts
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/icon.service.ts
@@ -21,7 +21,6 @@ export class IconService {
         private iconRegistry: MatIconRegistry) {
 
         this.icons.set('Account', ['account_balance.svg', this.defaultIconProvider]);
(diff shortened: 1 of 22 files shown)


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/icon.service.ts
```diff
@@ -21,6 +21,7 @@ export class IconService {
         private iconRegistry: MatIconRegistry) {
 
         this.icons.set('Account', ['account_balance.svg', this.defaultIconProvider]);
+        this.icons.set('AccountCircle', ['account_circle.svg', this.defaultIconProvider]);
```
dankaste, line 24: We actually don't need this anymore. It should work to just send over the icon key `account_circle`.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/sale/sale.component.ts
```diff
@@ -74,6 +74,10 @@ export class SaleComponent extends PosScreen<SaleInterface> {
         }
     }
 
+    public doMenuItemAction(menuItem: IActionItem) {
```
dankaste, line 77: you should be able to just call doAction from the html template

chenson42:
This needs to be configurable.   Another customer wants the original layout.

chenson42:
I think the button should be the same height prior to and after the custom is linked

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/accent-button/accent-button.component.scss
```diff
@@ -0,0 +1,31 @@
...
+  &.mobile {
+    border-width: 1px;
+    font-size: $text-xs-mobile*1.25;
+    min-width: 75px;
+    padding: $text-sm-mobile/8 $text-sm-mobile*.75;
+  }
+}
+
+:host ::ng-deep app-icon {
+  margin: -4px -4px -4px 0;
```
dankaste, line 29: if we can avoid ng-deep that would be best. Last I checked angular announced its plans to deprecate this.

dankaste (commented):
This is just personal preference but I like the button on the left similar to how it looks in the mobile view.

mmichalek:
This is a more precise mock per Chris' comment and showing the JM colors. 
image

image

chenson42:
> This is just personal preference but I like the button on the left similar to how it looks in the mobile view.

Yeah.  This was the compromise between Petco and AEO.

chenson42 (approved):

r-cochran:
latest change included making the Loyalty ID label i18n based instead of static

PR #1123, opened 2021-02-12
Bugfix/gs1-vli-validation

JJStarrett:
### Issues Fixed
 - [4907](https://petcoalm.atlassian.net/browse/PDPOS-4907)

### Summary
GS1 Coupons should now consider any value in the secondary and tertiary purchase company VLI greater than 6 as a flag to ignore the respective field. This should allow certain coupons with a "9" in those fields to work properly.

### DIFF
--- a/openpos-util/src/main/java/org/jumpmind/pos/coupons/GS1Coupon.java
+++ b/openpos-util/src/main/java/org/jumpmind/pos/coupons/GS1Coupon.java
@@ -105,8 +105,10 @@ public class GS1Coupon {
                         i += 3;
 
                         secondaryPurchaseCompanyPrefixVLI = Integer.parseInt(gs1Databar.substring(i, ++i));
-                        secondaryPurchaseCompanyPrefix = gs1Databar.substring(i, i + 6 + +secondaryPurchaseCompanyPrefixVLI);
-                        i += 6 + +secondaryPurchaseCompanyPrefixVLI;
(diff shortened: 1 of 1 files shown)


JJStarrett:
I want to kind of note how exactly I came to this decision, as I'm certainly not any expert on GS1 coupons. The root of the issue Petco was having is that they had a '9' in the secondary purchase company vli slot - which made the secondary purchase company field appear longer than the databar input to the parser. I did a little research into GS1, and it seems like that '9' is supposed to indicate that the secondary purchase company field is supposed to be skipped. And for the data bar `811020840243103274533001100001011000093210531` this would mean that the last digits `210531` are the date `2021-05-31` which seems to make sense. 

I was led to believe this as you'll note on [this site](https://www.databar-barcode.info/databar-coupon/coupon_data_examples/) several examples show the 9 implying an "N/A" value for the following secondary purchase company field:
image

My best guess as to why this is comes from [this site](https://www.databar-barcode.info/databar-coupon/coupon_data_examples/) which notes that the purchase company VLI's can only be 0 to 6. I gather that must mean any value greater than 6 implies that the subsequent purchase company field should be skipped:
image

Hopefully that makes sense, and if anyone has any additional clarification on this issue, I'd appreciate it.

jrmihalick:
Thanks for documenting this @JJStarrett !

jrmihalick (approved):

PR #1127, opened 2021-02-16
Prevent Multiple Clients to Connect to the Same Device Simultaneously

liambell-le:
### Issues Fixed
https://jira.ae.com/browse/PCOM-1331

### Summary
The current state of the session listeners allows for multiple clients to connect to the same device due to the nature of STOMP one to many. In the WebSocketConfig, check if a an existing session id is associated to a device token, and do not allow another connection until the association no longer exists. To allow the original client to reconnect on refresh, check that the session id is associated to the device token, remove the association during DISCONNECT stomp command. Once disconnected, on the CONNECT stomp, the new session id is added as an association to the device token.

### DIFF
--- a/openpos-server-core/src/main/java/org/jumpmind/pos/server/config/WebSocketConfig.java
+++ b/openpos-server-core/src/main/java/org/jumpmind/pos/server/config/WebSocketConfig.java
@@ -1,9 +1,14 @@
 package org.jumpmind.pos.server.config;
 
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.mutable.MutableBoolean;
+import org.jumpmind.pos.devices.DeviceNotAuthorizedException;
+import org.jumpmind.pos.server.model.SessionContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


chenson42 (approved):

openpos-server-core/src/main/java/org/jumpmind/pos/server/config/WebSocketConfig.java
```diff
@@ -77,6 +85,44 @@ public void configureMessageBroker(MessageBrokerRegistry config) {
...
+
+                    for (Map.Entry<String, SessionContext> pair : deviceToSessionMap.entrySet()) {
+                        if (pair.getValue().getSessionId().equals(sessionId)) {
+                            isDeviceConnected = true;
+                        }
+                    }
+
+                    if (!isDeviceConnected) {
+                        throw new DeviceNotAuthorizedException();
+                    }
```
kdelorey, line 113: nit: you could early out this if you wanted to. Unlikely to save much because I imagine there won't be a high quantity of clients at any given time, but:

```suggestion
                    for (Map.Entry<String, SessionContext> pair : deviceToSessionMap.entrySet()) {
                        if (pair.getValue().getSessionId().equals(sessionId)) {
                            throw new DeviceNotAuthorizedException();
                        }
                    }
```

kdelorey (approved):
lgtm

dankaste (approved):

PR #1139, opened 2021-02-18
Round up donation at tendering

JJStarrett:
### Issues Fixed
 - [4220](https://petcoalm.atlassian.net/browse/PDPOS-4220)

### Summary
Adds UI support for an optional "Round Up" donation button to the tender options screen. Most of the buttons functionality (aside from the checkmark icon) is driven from the server, so in the future the button's functionality could easily be altered on the server without editing the client.

### Screenshots
https://user-images.githubusercontent.com/39058176/108403310-90c54500-71ec-11eb-9753-951824c1879a.mp4


blubaugha (commented):
This looks nice @JJStarrett , just a couple comments:

1. I think you should move the icon to the left of the text because that's where other icons are placed in the app
2. I'm curious why this round-up donation button was added here. Mainly because there's an option to round up on the **Charity Donation Dialog** right before tendering:
Screen Shot 2021-02-18 at 4 35 10 PM

JJStarrett:
> 1. I think you should move the icon to the left of the text because that's where other icons are placed in the app

Good point, adjusted: 
image

> 2. I'm curious why this round-up donation button was added here. Mainly because there's an option to round up on the **Charity Donation Dialog** right before tendering:

This issue has been in my queue for quite a long time now, and I think this was the reason why. I think Petco specifically requested this - JDA has functionality similar to this where the round up option is separated and stackable with other donations. And, I think that's the primary reason.

JJStarrett:
> > 1. I think you should move the icon to the left of the text because that's where other icons are placed in the app
> 
> Good point, adjusted:

So after thinking about this a little more, I'm wondering if it does actually make sense to put it on the right. Most Option Items in the app have their icons on the **left** - which move you forward in a flow - while items that "do something" to the current screen actually trend towards having their icons on the **right**, like "remove" and "options" on the sale screen:

image
image

Perhaps it would make sense to put Round Up's icon on the **right** since it doesn't move the user through tendering, it just alters the current state. Though, it would be the only button in tendering with its icon on the right then. @mmichalek I'd be curious for your thoughts on this

stojankz (commented):
my only comment is that I think the icon for this button should match the icon of the Round Up button from the donations screen. Either change this one to the + or the one on the Donation screen to this one for consistency

jrmihalick:
> my only comment is that I think the icon for this button should match the icon of the Round Up button from the donations screen. Either change this one to the + or the one on the Donation screen to this one for consistency

I agree with @stojankz .  I think use the '+' icon since it's more indicative of the button function.

@blubaugha Petco specifically asked for this feature because some customer want to both round up AND donate a fixed amount through the PIN Pad or when the cashier asks them.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/tender-part/tender-part.component.ts
```diff
@@ -26,6 +27,21 @@ export class TenderPartComponent extends ScreenPartComponent<TenderPartInterface
         else {
             this.amountCss = '';
         }
+        if (this.screenData.roundUpAvailable)
```
jrmihalick, line 30: this.isRoundUpAvailable = this.screenData.roundUpAvailable

jrmihalick (changes requested):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/tender-part/tender-part.component.ts
JJStarrett: Adjusted.

JJStarrett:
New button:

image

mmichalek:
How about this one?
image

It's from https://materialdesignicons.com/ - called heart-plus-outline

JJStarrett:
> How about this one?
> image
> 
> It's from https://materialdesignicons.com/ - called heart-plus-outline

Oh, I like this one. I think it communicates rounding up is a charity donation as well as increasing your current balance. It looks like our icon.service.ts can't find the heart_plus_outline.svg. Is there a way to add new icons to the project?

JJStarrett:
image

I would change the round up button on the prompt to this icon as well if we made this change. Thoughts?

jrmihalick:
> I would change the round up button on the prompt to this icon as well if we made this change. Thoughts?

I agree.

jrmihalick (approved):

stojankz (approved):

PR #1142, opened 2021-02-19
Add X-button to the newly styled status screen

apomanytskyi:
Add X-button to the newly styled status screen

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/status/status-details/status-details.component.html
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/status/status-details/status-details.component.html
@@ -3,6 +3,15 @@
         <div class="header" gdArea="header" gdColumns="auto 1fr" responsive-class>
             <span class="x x-first x-last">{{'status:details.dialog.title.devicepairing' | localize}}</span>
         </div>
+        <app-icon-button
+                class="back-button"
+                gdArea="backbutton"
+                [iconName]="'close'"
+                (click)="close()"
+        ></app-icon-button>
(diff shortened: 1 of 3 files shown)


apomanytskyi:
Screen Shot 2021-02-19 at 10_Fotor

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/status/status-details/status-details.component.html
```diff
@@ -3,6 +3,15 @@
...
             <span class="x x-first x-last">{{'status:details.dialog.title.devicepairing' | localize}}</span>
         </div>
+        <app-icon-button
+                class="back-button"
+                gdArea="backbutton"
+                [iconName]="'close'"
+                (click)="close()"
+        ></app-icon-button>
+        <mat-form-field class="invisible">
+            <input matInput cdkFocusInitial type="text" responsive-class/>
```
apomanytskyi, line 13: In most popups we have inputs with autofocus (any search or data entry) or we also have no X-button (for example, popup when you click "Options" button on some item from sale transaction).
So here if I just add X-button - I always have focus on it... That's why I decide to add hidden input and set focus on it... If someone knows a better solution - I'll be grateful...

kdelorey: I think the hidden input is typically the solution for this problem. I also think the hidden input is done elsewhere. @dankaste might know better though.

kdelorey (approved):
lgtm

PR #1143, opened 2021-02-19
Allow username entry when current user null

jchirico8:
### Issues Fixed
[PDPOS-4986](https://petcoalm.atlassian.net/browse/PDPOS-4986)

### Summary
Toggle the username field on the lock screen based on data.userText. This is so when the currentUser is null, we can still unlock the screen. Also made the 2 submit actions be consistent in the data format they sent off.

### Screenshots
image


jrmihalick (approved):

jchirico8:
Was able to properly reproduce the bug and fixed it in a different manner in https://github.com/JumpMind/commerce/pull/3151

PR #1151, opened 2021-02-24
Feature/record sql statement values in data access violation error

JJStarrett:
### Issues Fixed
 - Logging to help diagnose [5009](https://petcoalm.atlassian.net/browse/PDPOS-5009)

### Summary
Adds logging for data violations exceptions when performing insert, update, or delete statements and records the values trying to be altered. Also adds a small bit of warning logging to note when tag values are inserted with an empty value instead of the default '*' or some other non-empty value. Both are set at the **warning** level as they should not often happen.

### Screenshots
image

### DIFF
--- a/openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
+++ b/openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
@@ -550,7 +550,7 @@ public class DBSession {
             message = message.substring(0, message.indexOf("SQL statement"));
         }
         errorStatement.append(": " + message);
-        log.warn(errorStatement.toString());
+        throw new DuplicateKeyException(errorStatement.toString(), e);
     }
 
     protected void batchInternal(List<? extends  AbstractModel> models, DmlType dmlType) {
--- a/openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
+++ b/openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
@@ -541,9 +541,9 @@ public class DBSession {
     protected void logDMLError(DmlType type, String sql, Object[] values, DataAccessException e) {
         StringBuilder errorStatement = new StringBuilder();
         errorStatement.append("There was a data access violation in " + type + " statement: ");
-        for (Object value : values) {
-            sql = sql.replaceFirst("[?]", String.valueOf(value));
-        }
+        LogSqlBuilder builder = new LogSqlBuilder();
+        Object[] rawArgs = cleanArgs(values);
+        sql = builder.buildDynamicSqlForLog(sql, rawArgs, null);
         errorStatement.append(sql);
         String message = e.getCause().getMessage();
         if (message.contains("SQL statement")) {
--- a/openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
+++ b/openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
@@ -534,21 +534,23 @@ public class DBSession {
             return jdbcTemplate.getJdbcOperations().update(sql, values, types);
         } catch (DataAccessException e) {
             logDMLError(type, sql, values, e);
-            return 0;
+            return jdbcTemplate.getJdbcOperations().update(sql, values, types);
         }
     }
 
     protected void logDMLError(DmlType type, String sql, Object[] values, DataAccessException e) {
         StringBuilder errorStatement = new StringBuilder();
-        errorStatement.append("There was an error with " + type + " statement: ");
-        errorStatement.append(sql.substring(0, sql.indexOf("(?")));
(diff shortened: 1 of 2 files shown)


jrmihalick (approved):

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
```diff
@@ -530,7 +530,27 @@ protected int excecuteDml(DmlType type, ModelWrapper model, Table table) {
...
+        for (Object value : values) {
+            sql = sql.replaceFirst("[?]", String.valueOf(value));
+        }
+        errorStatement.append(sql);
+        String message = e.getCause().getMessage();
+        if (message.contains("SQL statement")) {
+            message = message.substring(0, message.indexOf("SQL statement"));
+        }
+        errorStatement.append(": " + message);
+        log.warn(errorStatement.toString());
```
mmichalek, line 553: This should be the message of a re-thrown exception.   Like:
throw new PersistException("SQL Statement caused exception: " + statement, ex);

I think it's important to keep logging consolidated to one entry per error as much as possible. Otherwise, in a highly concurrent environment, you can end up with a single error quite spread out in the log.

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
```diff
@@ -530,7 +530,27 @@ protected int excecuteDml(DmlType type, ModelWrapper model, Table table) {
...
+        } catch (DataAccessException e) {
+            logDMLError(type, sql, values, e);
+            return jdbcTemplate.getJdbcOperations().update(sql, values, types);
+        }
+    }
+
+    protected void logDMLError(DmlType type, String sql, Object[] values, DataAccessException e) {
+        StringBuilder errorStatement = new StringBuilder();
+        errorStatement.append("There was a data access violation in " + type + " statement: ");
+        for (Object value : values) {
```
mmichalek, line 544: We should use LogSqlBuilder.buildDynamicSqlForLog(...) for format the SQL statment for the log.

mmichalek (changes requested):

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
JJStarrett: I was actually initially re-throwing a PersistException here, but I noticed some insert statements that would normally fail over to an update, would not, and the app would break where it probably shouldn't. I could re-throw the DataAccessViolation Exception and catch it where the save() method in DBSession.java catches DuplicateKey Exceptions:

```
                try {
                    insert(model, table);
                } catch (DuplicateKeyException ex) {
                    log.info("Insert of entity failed, failing over to an update: " + argModel);
                    int updateCount = update(model, table);
                    if (updateCount < 1) {
                        throw new PersistException("Failed to perform an insert or update on entity. Do the DB primary key and unique fields "
                                + "match what's understood by the code?  " + argModel, ex);
                    }
                }
```

Or is there a different exception I might be able to throw that would not halt the application?

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
JJStarrett: Adjusted. Thanks for the heads-up, Mark

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
mmichalek: Ah - it should just check and re-throw the DuplicateKeyException actually.

JJStarrett: Gotcha, adjusted.

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSession.java
```diff
@@ -530,7 +530,27 @@ protected int excecuteDml(DmlType type, ModelWrapper model, Table table) {
...
+        LogSqlBuilder builder = new LogSqlBuilder();
+        Object[] rawArgs = cleanArgs(values);
+        sql = builder.buildDynamicSqlForLog(sql, rawArgs, null);
+        errorStatement.append(sql);
+        String message = e.getCause().getMessage();
+        if (message.contains("SQL statement")) {
+            message = message.substring(0, message.indexOf("SQL statement"));
+        }
+        errorStatement.append(": " + message);
+        throw new DuplicateKeyException(errorStatement.toString(), e);
```
mmichalek, line 553: Maybe I'm missing something - but this method says it will "log" (logDMLError) but actually throws an exception . I'm also not seeing how we know to throw a DuplicateKeyException the way this is currently.  Let's chat about it tomorrow @JJStarrett

PR #1160, opened 2021-03-01
loyalty button on high pixel-ratio devices

r-cochran:
### Summary
Fixed a styling issue for the Loyalty button when on high pixel-ratio devices that caused the typography to increase the font size instead of get smaller on widths in the 1175 to 840 range.

### Screenshots
Screenshots go here.
image

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/sale-total-panel/sale-total-panel.component.scss
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/sale-total-panel/sale-total-panel.component.scss
@@ -19,7 +19,7 @@
         font-size: 36px;
     }
     .link-customer {
-        height: 102px;
+        min-height: 102px;
         @extend %text-md;
         border-width: 2px;
         padding: $text-md/2 $text-md*1.25;


dankaste:
Should we maybe just not make the text so big here? Personally I think it looks odd to have 4 lines of text in the button.

r-cochran:
> Should we maybe just not make the text so big here? Personally I think it looks odd to have 4 lines of text in the button.

The text resizing is an inherited behavior from the `@extend %text-md;` and its use of rems for the text block. We mirrored the styling of the old loyalty button and the checkout button which used the same style extension for the typography. 

On a mac with the display settings set to Scaled and More Space (1920x1200) is when the situation occurs, similar text changes occur with the Checkout button on the bottom of the page.  Is this a resolution that is supported? If so, there are probably more places in the system that have similar issues that use the `@extend %text-md;`.

The checkout button also appears to get a little weird when in this same resolution/browser width. 
image

dankaste:
Yeah that makes sense. But this isn't the typical. Most buttons have one maybe 2 words not 5 so we might just need to enhance the look and feel here. It looks like the way the totals bar layout was solved is messing up the checkout button. looks like it is stretching it instead of aligning it.

r-cochran:
Changes from a thread involving the request for icons/text for a client have lead to other/different changes to be made. Closing this PR.

PR #1167, opened 2021-03-02
Capacitor Storage

kdelorey:
### Summary
The personalization service has been updated to shim in capacitor's storage plugin when running natively on iOS (or technically android too). This should resolve an issue where updates to the application were destroying personalization data.

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/personalization/personalization.service.ts
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/personalization/personalization.service.ts
@@ -1,10 +1,12 @@
-import { Injectable, Injector } from '@angular/core';
+import { Injectable } from '@angular/core';
 import { HttpClient } from '@angular/common/http';
 import { PersonalizationConfigResponse } from './personalization-config-response.interface';
-import { Observable, BehaviorSubject, throwError } from 'rxjs';
+import { Observable, BehaviorSubject, throwError, Subject, zip } from 'rxjs';
 import { catchError, map, tap } from 'rxjs/operators';
 import { PersonalizationRequest } from './personalization-request';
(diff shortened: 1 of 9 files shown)


chenson42 (approved):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/personalization/personalization.service.ts
```diff
@@ -189,13 +202,59 @@ export class PersonalizationService {
...
+        PersonalizationService.setStorageValue('deviceToken', token);
         this.deviceToken$.next(token);
     }
 
     public refreshApp() {
         window.location.reload();
     }
-}
 
+    private static canUseCapacitorStorage(): boolean {
```
dankaste, line 213: wondering if we should add storage through a plug-in or separate service. Then it can default to localStorage or use a native plugin if present?

kdelorey: I think that'd be ideal. In slack the other day @chenson42 showed desire in just having it done the quick and dirty way likely because it was desirable to have this asap since it was going to affect SIT pretty much every day.

I assumed we could revisit this to clean it up after its in operation. I don't think it'd be hard to abstract this away into a plug-able service at all.

dankaste: maybe for now just moving it to a separate service quickly so we don't bring platform logic into the personalization service

kdelorey: I've made storage pluggable, check it out!

dankaste (approved):
thanks

PR #1174, opened 2021-03-05
Bugfix/keybindings not applying

jchirico8:
### Issues Fixed
[PDPOS-4839](https://petcoalm.atlassian.net/browse/PDPOS-4839)

### Summary
An instantiation of ActionItem was switched from constructor to lombok builder. This wiped out the default value for these fields and would require them to be declared when building the action item.
From the lombok documentation:
```
If a certain field/parameter is never set during a build session, then it always gets 0 / null / false. 
If you've put @Builder on a class (and not a method or constructor) you can instead specify the 
default directly on the field, and annotate the field with @Builder.Default
```
Another note is that this new annotation will block the default assignment from explicitly defined constructors. To get past that, I'm calling the lombok generated NoArgsConstructor from our defined ones, which handles any Builder.Default assignments.

### DIFF
--- a/openpos-flow/src/main/java/org/jumpmind/pos/core/ui/ActionItem.java
+++ b/openpos-flow/src/main/java/org/jumpmind/pos/core/ui/ActionItem.java
@@ -44,7 +44,8 @@ public class ActionItem implements Serializable {
     protected String defaultPayload;
 
     protected String icon;
-    @Builder.Default protected boolean enabled = true;
+    @Builder.Default
+    protected boolean enabled = true;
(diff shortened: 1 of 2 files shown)


mmichalek (approved):

jrmihalick:
I think this warrants a unit test so that someone doesn't accidentally break this in the future.

jrmihalick (changes requested):
Let's add a unit test to cover this.

jchirico8:
> Let's add a unit test to cover this.

I added 13

jrmihalick (approved):

PR #1176, opened 2021-03-09
Added config to allow manual entry of receipt detail. Disabled by default.

apomanytskyi:
Added logic and config to allow manual entry of receipt detail. Disabled by default.
To enable - set openpos.returns.allowManualEntryOfReceiptDetails=true
allowRecreateTrans


openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/return/return-trans-details/return-trans-details-dialog.component.scss
```diff
@@ -28,9 +29,21 @@
...
+    justify-content: right;
+    grid-auto-flow: column;
 }
 
 .mat-dialog-content {
     padding: 0;
     margin: 0;
 }
+
+::ng-deep {
```
stevencarley, line 43: change to :host ::ng-deep to prevent style from possibly leaking to other components.

stevencarley (dismissed):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/return/add-item-to-return-transaction/add-item-to-return-transaction-dialog.component.html
```diff
@@ -0,0 +1,4 @@
+<app-dialog-header></app-dialog-header>
```
dankaste, line 1: Did just using the existing dynamic for screen not work here?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/return/return-trans-details/return-trans-details-dialog.component.html
```diff
@@ -14,7 +14,19 @@
                 </div>
             </mat-dialog-content>
         </app-content-card>
+        <ng-container *ngIf="employeeTransaction">
+            <mat-checkbox id="{{employeeTransaction.id}}"
+                          [checked]="employeeTransaction.checked"
+                          (change)="onEmployeeTransaction($event)" [value]="employeeTransaction.value"
+                          [disabled]="employeeTransaction.readOnly" [labelPosition]="employeeTransaction.labelPosition"
```
dankaste, line 21: I don't think we need disabled here. In the case where its not used we just won't add it to the screen. Also don't send over the labelPosition. The server shouldn't really care about layout decisions

dankaste (changes requested):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/return/return-trans-details/return-trans-details-dialog.component.html
```diff
@@ -14,7 +14,19 @@
                 </div>
             </mat-dialog-content>
         </app-content-card>
+        <ng-container *ngIf="employeeTransaction">
+            <mat-checkbox id="{{employeeTransaction.id}}"
+                          [checked]="employeeTransaction.checked"
+                          (change)="onEmployeeTransaction($event)" [value]="employeeTransaction.value"
+                          [disabled]="employeeTransaction.readOnly" labelPosition="'before'"
```
dankaste, line 21: Does labelPosition="'before'" work? I think if you take the [] off labelPosition you don't need the single ticks '' anymore

dankaste (approved):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/return/return-trans-details/return-trans-details-dialog.component.html
apomanytskyi: Yes, you are right, thanks!

dankaste (approved):

PR #1181, opened 2021-03-10
Match default implementations for endpoint overrides

eddielegg-le:
### Summary
Match override endpoints with matching implementation or "default"

### DIFF
--- a/openpos-service/src/main/java/org/jumpmind/pos/service/EndpointInvoker.java
+++ b/openpos-service/src/main/java/org/jumpmind/pos/service/EndpointInvoker.java
@@ -189,14 +189,23 @@ public class EndpointInvoker implements InvocationHandler {
     protected Object findBestEndpointOverrideMatch(String path, String implementation, Collection<Object> endpointOverrides) {
         Object bestMatch = null;
         List<SimpleEntry<Object,EndpointOverride>> pathMatchedOverrides = endpointOverrides.stream()
-                .map(o -> new SimpleEntry<Object,EndpointOverride>(o, ClassUtils.resolveAnnotation(EndpointOverride.class, o)))
+                .map(o -> new SimpleEntry<>(o, ClassUtils.resolveAnnotation(EndpointOverride.class, o)))
                 .filter(entry -> entry.getValue().path().equals(path)
                 ).collect(Collectors.toList());
 
         if (pathMatchedOverrides.size() > 0) {
-            List<SimpleEntry<Object, EndpointOverride>> implMatchedOverrides = pathMatchedOverrides.stream()
-                    .filter(entry -> entry.getValue().implementation().equals(Endpoint.IMPLEMENTATION_DEFAULT) || entry.getValue().implementation().equals(implementation))
-                    .collect(Collectors.toList());
+            List<SimpleEntry<Object, EndpointOverride>> implMatchedOverrides;
+            if(pathMatchedOverrides.stream().anyMatch(entry -> entry.getValue().implementation().equals(implementation))) {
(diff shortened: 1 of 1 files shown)


eddielegg-le:
Found this issue when an overridden endpoint implementation wasn't specified but did have an implementation configured in application.yml. It looks like it was this way before the training mode impl...was this an intentional change?

https://github.com/JumpMind/openpos-framework/pull/1094/files#diff-89db084bc4e2c85433a84ef3590dba6133a88e8b274f3040c307fef944be49f9L143

JoeBrownlee (commented):
I had the exact same thing you do, Eddie, at one point. There was a reason that someone wanted it reverted back to the original version of line 198. I think it was because it was causing a bunch of error messages on startup for Petco. @mmichalek might be a good one to weigh in on this.

jrmihalick:
@eddielegg-le Are you saying that it was implemented in the same way as your change before bringing over the EndpointInvoker class from master?  That is, it was implemented the same way as your change in the `EndpointDispatchInvocationHandler.java` class? (https://github.com/JumpMind/openpos-framework/pull/1094/files#diff-89db084bc4e2c85433a84ef3590dba6133a88e8b274f3040c307fef944be49f9)

eddielegg-le:
> @eddielegg-le Are you saying that it was implemented in the same way as your change before bringing over the EndpointInvoker class from master? That is, it was implemented the same way as your change in the `EndpointDispatchInvocationHandler.java` class? (https://github.com/JumpMind/openpos-framework/pull/1094/files#diff-89db084bc4e2c85433a84ef3590dba6133a88e8b274f3040c307fef944be49f9)
> 
> For some reason, that link you provided doesn't seem to be working for me, but this link above takes me to the deleted `EndpointDispatchInvocationHandler.java` class.

Bleh, yeah...I think it's because the diff is collapsed. Here's the old implementation before bringing it over:
```
                List<SimpleEntry<Object, EndpointOverride>> defaultImplOverrides = pathMatchedOverrides.stream()
                        .filter(entry -> EndpointOverride.IMPLEMENTATION_DEFAULT.equalsIgnoreCase(entry.getValue().implementation()) ||
                                StringUtils.isBlank(entry.getValue().implementation()))
```

mmichalek:
@eddielegg-le  @jrmihalick 
What issue is this fixing?

Are you saying that if you have a base endpoint: 
`"/authorize" implemention=adyan`
and you have an override of:
`"/authorize" implementation=default override=true`
that it should or should not override the base?  Is that the question that's being asked here?

jrmihalick:
@mmichalek Let's discuss this tomorrow, post standup.  After reviewing this change and a related change in Petco, we have a problem with endpoint overriding that we need to fix.

eddielegg-le:
> @eddielegg-le @jrmihalick
> What issue is this fixing?
> 
> Are you saying that if you have a base endpoint:
> `"/authorize" implemention=adyan`
> and you have an override of:
> `"/authorize" implementation=default override=true`
> that it should or should not override the base? Is that the question that's being asked here?

This was addressing an issue where the controller implementation was defined in the yml and the base endpoint was overridden, but the endpoint implementation didn't match the yml implmentation:

```
# application-corp.yml
openpos:
...
  services:
    returnsmgmt_1_1:
       implementation: appriss



@Endpoint(path="/returnsmgmt/version/1_1/search")
@EndpointOverride(path="/returnsmgmt/version/1_1/search")
```
In this scenario, the EndpointInvoker matched the RestController with the original endpoint instead of the EndpointOverride

jrmihalick (approved):

PR #1183, opened 2021-03-11
Loyalty Member Name and Member ID get cut off

apomanytskyi:
JMC Mobile - Loyalty Member Name and Member ID get cut off
We always had font size 24px. In browser we can see that the text does not fit.
On a mobile device, the surname is just truncated.
1938_problem
Suggestion - use dynamic font size for loyalty button.
1938_fixed

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/sale-total-panel/sale-total-panel.component.scss
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/sale-total-panel/sale-total-panel.component.scss
@@ -46,7 +46,7 @@
                 .name {
                     grid-area: Name;
                     .customer-name {
-                        font-size: 24px;
+                        font-size: 1.5vw;
                         .loyalty-keybind {
                             margin-left: 10px;
                         }
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/sale-total-panel/sale-total-panel.component.scss
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/sale-total-panel/sale-total-panel.component.scss
@@ -46,7 +46,7 @@
                 .name {
                     grid-area: Name;
                     .customer-name {
-                        font-size: 1.5vw;
+                        @extend %text-md;
                         .loyalty-keybind {
                             margin-left: 10px;
                         }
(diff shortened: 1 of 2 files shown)


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/sale-total-panel/sale-total-panel.component.scss
```diff
@@ -46,7 +46,7 @@
                 .name {
                     grid-area: Name;
                     .customer-name {
-                        font-size: 24px;
+                        font-size: 1.5vw;
```
dankaste, line 49: would be better to do this with our responsive classes and font sizes. You should be able to see evidence of this elsewhere in the codebase. look for usage of the response-class directive

dankaste (approved):

dankaste (changes requested):
use responsive font sizes instead of vw

apomanytskyi:
@dankaste
Yes, but with the responsive-class there still will be a bug in some cases ... It all depends on how long the first name + last name will be ...
Because this is the dynamics length data - the only solution I can see - a fully dynamic text size calculation.
The default size is now 24px. If use %text-md, the bug remains.
If change it to %text-sm - the bug will be fixed, but no one asked to make the text smaller by default.
If you have any other suggestions, it will be cool... because now I'm not sure this is the best possible solution...
1928_fix_with_responsive

dankaste:
@apomanytskyi what if we just allow wrapping in the super long case and I'd just make it smaller I don't think it needs to be 24px

apomanytskyi:
Ok, how about this: I reduced the default font-size to 18px. Now the bug will be fixed, because first and last name from test-case will fit.
For super long cases - the text will be hidden outside the borders of the button, but will be available on scrolling.
I do not think that this one will be used often and will appear, but I would like to leave the ability to fully view the first name and last name, regardless of its length.
Perhaps I can try to split the text into multiple lines if needed, but I'm afraid the other styles might break.
Or I can just cut the text by "text-overflow" but I like this solution less ...

1928_with_scroll

dankaste:
> Ok, how about this: I reduced the default font-size to 18px. Now the bug will be fixed, because first and last name from test-case will fit.
> For super long cases - the text will be hidden outside the borders of the button, but will be available on scrolling.
> I do not think that this one will be used often and will appear, but I would like to leave the ability to fully view the first name and last name, regardless of its length.
> Perhaps I can try to split the text into multiple lines if needed, but I'm afraid the other styles might break.
> Or I can just cut the text by "text-overflow" but I like this solution less ...
> 
> 1928_with_scroll

I think I'd rather wrap than scroll. It really shouldn't ever be the case that its that long anyways. this is just incase.

apomanytskyi:
text wrapping
1928_text_wrap

dankaste (approved):

PR #1185, opened 2021-03-12
Match default implementations for endpoint overrides (#1181)

eddielegg-le:
### Summary
Match override endpoints with matching implementation or "default"

This was addressing an issue where the controller implementation was defined in the yml and the base endpoint was overridden, but the endpoint implementation didn't match the yml implmentation:

```
# application-corp.yml
openpos:
...
  services:
    returnsmgmt_1_1:
       implementation: appriss

@Endpoint(path="/returnsmgmt/version/1_1/search")
@EndpointOverride(path="/returnsmgmt/version/1_1/search")
```
In this scenario, the EndpointInvoker matched the RestController with the original endpoint instead of the EndpointOverride. The


eddielegg-le:
> Does this need to be merged still?

I guess I'm not sure based on the conversation in the linked PR:

https://github.com/JumpMind/openpos-framework/pull/1263#issuecomment-819602666

chenson42:
Just gonna close this.   We can revisit in the future if need be.

PR #1187, opened 2021-03-16
Use DB platform to prefix all table attributes with shadow table prefix

eddielegg-le:
### Summary
Adding training mode table indices was causing conflicts with some older indices that were not prefixed with the module name. Updated to ensure that all attributes for shadow tables are prefixed correctly using the DB platform specified.


jrmihalick:
I'll pass to @mmichalek et al on this.

mmichalek (approved):

jrmihalick:
@eddielegg-le Anything keeping you from merging?  If not, could you merge to 1.10 instead?

eddielegg-le:
> @eddielegg-le Anything keeping you from merging? If not, could you merge to 1.10 instead?

I forgot it even existed 😆  Sure thing, I'll move to 1.10 and test there then merge

PR #1196, opened 2021-03-19
Make timers on sale screen appear without delay

levshunovm:
### Summary
Make timers on the sale screen appear without delay


dankaste (changes requested):
See my comment on the other PR, but I don't think this is the issue they are seeing.

PR #1197, opened 2021-03-19
Make timers on sale screen appear without delay

levshunovm:
### Summary
Make timers on the sale screen appear without delay


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/status-strip/status-strip.component.ts
```diff
@@ -15,8 +15,8 @@ import { ScreenPart } from '../../decorators/screen-part.decorator';
 })
 export class StatusStripComponent extends ScreenPartComponent<StatusStripInterface> {
 
-    date = interval(1000).pipe( map( () => Date.now()));
-    timer = interval(1000).pipe(map( () => {
+    date = timer(0, 1000).pipe(map( () => Date.now()));
```
dankaste, line 18: this isn't the delay they are seeing. The delay they see is much longer and I wasn't able to reproduce locally with an iPad. You'll need to work with QE to trouble shoot this, it may not even be related to the IPAD at all.

dankaste (changes requested):

PR #1202, opened 2021-03-22
Subscribe to the "Print" message on transaction details

jchirico8:
### Summary
[PDPOS-5145](https://petcoalm.atlassian.net/browse/PDPOS-5145)
Adding a subscription to the transaction details component to listen for the print message and open the browser print dialog. The POS TransactionDetailsState does not send this message and will not be affected by this change even though this component is shared with CentralTransactionDetailsState

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/transaction-details/transaction-details.component.ts
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/transaction-details/transaction-details.component.ts
@@ -4,6 +4,9 @@ import { ScreenComponent } from '../../shared/decorators/screen-component.decora
 import { PosScreen } from '../pos-screen/pos-screen.component';
 import { OpenposMediaService, MediaBreakpoints } from '../../core/media/openpos-media.service';
 import { Observable } from 'rxjs';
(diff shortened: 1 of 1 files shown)


openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/transaction-details/transaction-details.component.ts
```diff
@@ -27,8 +34,14 @@ export class TransactionDetailsComponent extends PosScreen<TransactionDetailsInt
       [MediaBreakpoints.DESKTOP_PORTRAIT, false],
       [MediaBreakpoints.DESKTOP_LANDSCAPE, false]
     ]));
+    this.sessionService.getMessages(MessageTypes.PRINT).subscribe((message) => {
```
jrmihalick, line 37: Should this component unsubscribe from the print messages in its destroy?  I'm not sure, just checking.

jchirico8: I may not need this change at all anymore. I'm testing what I think is the proper way to send print messages now

jchirico8:
This is no longer necessary.

PR #1208, opened 2021-03-24
Create AddressUIMessage that AutoCompleteAddressUIMessage extends from. PCOM-2289

liambell-le:
### Summary
Abstract the properties of AutoCompleteAddressUIMessage to a generic AddressUIMessage that will not implement the IHasAutoComplete interface. This change will allow for address forms that do not utilize the google api auto complete functionality.

### DIFF
--- a/openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/AddressUIMessage.java
+++ b/openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/AddressUIMessage.java
@@ -18,7 +18,6 @@ public class AddressUIMessage extends DynamicFormUIMessage {
     private StatusStripPart statusStrip = new StatusStripPart();
 
     public AddressUIMessage() {
-        setScreenType(UIMessageType.DYNAMIC_FORM);
         ActionItem submitButton = new ActionItem("Next", "Next", IconType.Forward);
         submitButton.setKeybind(KeyConstants.KEY_ENTER);
         setSubmitButton(submitButton);
--- a/openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/AddressUIMessage.java
+++ b/openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/AddressUIMessage.java
@@ -24,10 +24,8 @@ public class AddressUIMessage extends DynamicFormUIMessage {
     }
 
     public AddressUIMessage(String screenType) {
+        this();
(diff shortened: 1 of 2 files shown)


openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/AddressUIMessage.java
```diff
@@ -0,0 +1,75 @@
...
+public class AddressUIMessage extends DynamicFormUIMessage {
+
+    private static final long serialVersionUID = 1L;
+
+    private BaconStripPart baconStrip = new BaconStripPart();
+    private List<ActionItem> sausageLinks = new ArrayList<>();
+    private StatusStripPart statusStrip = new StatusStripPart();
+
+    public AddressUIMessage() {
+        setScreenType(UIMessageType.DYNAMIC_FORM);
```
stevencarley, line 21: Can change this to just call AddressUIMessage(UIMessageType.DYNAMIC_FORM)

stevencarley (changes requested):

openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/AddressUIMessage.java
liambell-le: Removed from the AddressUIMessage. Makes sense it's already set in DynamicFormUIMessage

openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/AddressUIMessage.java
```diff
@@ -0,0 +1,74 @@
...
+    private List<ActionItem> sausageLinks = new ArrayList<>();
+    private StatusStripPart statusStrip = new StatusStripPart();
+
+    public AddressUIMessage() {
+        ActionItem submitButton = new ActionItem("Next", "Next", IconType.Forward);
+        submitButton.setKeybind(KeyConstants.KEY_ENTER);
+        setSubmitButton(submitButton);
+    }
+
+    public AddressUIMessage(String screenType) {
```
stevencarley, line 26: Can change the constructor to
```
public AddressUIMessage(String screenType) {
    this();
    setScreenType(screenType);
}
```

stevencarley (changes requested):

stevencarley (approved):

PR #1212, opened 2021-03-26
Zip Code Validation

liambell-le:
### Summary
Added a new postal code input type to allow validation for either/or US/CA postal codes.


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/formatters/postal-code-generic.formatter.ts
```diff
@@ -0,0 +1,58 @@
...
+        /[0-9]/
+    ];
+    allowKey(key: string, newValue: string): boolean {
+        const len = newValue.length;
+
+        return PostalCodeGenericFormatter.FILTER_REGEXS_CA[len - 1].test(newValue[len - 1]) ||
+            PostalCodeGenericFormatter.FILTER_REGEXS_US[len - 1].test(newValue[len - 1 ]);
+    }
+
+    formatValue(value: string): string {
```
kdelorey, line 26: Does this like coerce the text as you're typing or something?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/formatters/postal-code-generic.formatter.ts
```diff
@@ -0,0 +1,58 @@
...
+export class PostalCodeGenericFormatter implements IFormatter {
+    static readonly FILTER_REGEXS_CA = [
+        /[abceghjklmnprstvxyABCEGHJKLMNPRSTVXY]/,
+        /[0-9]/,
+        /[abceghjklmnprstvwxyzABCEGHJKLMNPRSTVWXYZ]/,
+        /[0-9]/,
+        /[abceghjklmnprstvwxyzABCEGHJKLMNPRSTVWXYZ]/,
+        /[0-9]/
+    ];
+    static readonly FILTER_REGEXS_US = [
```
kdelorey, line 12: Do you need to worry about allowing valid delivery route numbers? (eg.12345-1234)

liambell-le: The requirement in the epic does not specify route. "Zip 5 (US) Postal code 6 (Canada with the a1a 1a1 format)".

mmichalek: I think we should avoid putting this validation burden on the client if possible.  My concern is that we would have to modify the angular app to take it international. Could the server specify the valid pattern(s) that the client could use?

stevencarley (approved):

PR #1216, opened 2021-03-29
Add support for training mode in DML queries

jrmihalick:
### Issues Fixed

**PDPOS-5307** (_[NGPOS][Training Mode] - Return Txn# in Training Mode is in sequence with Regular Mode txn# and Customer Receipt does NOT indicate Training Mode on return transaction receipt_)

### Summary
- Adds core support for necessary to fix training mode problems for any functional areas that use DML queries.  For example, updating the quantity available for return on a line item of an original transaction.


JoeBrownlee (commented):
Like the abstract superclass.

mmichalek (approved):

PR #1228, opened 2021-04-02
Changed the print exception to allow for status codes to be applied. …

eddielegg-le:
### Summary
Cherry pick commit to support warnings for printer out of paper exceptions

### DIFF
--- a/openpos-peripheral/src/main/java/org/jumpmind/pos/print/PrintException.java
+++ b/openpos-peripheral/src/main/java/org/jumpmind/pos/print/PrintException.java
@@ -1,28 +1,73 @@
 package org.jumpmind.pos.print;
 
 import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+import java.util.EnumSet;
 
 @Data
+@EqualsAndHashCode(callSuper = true)
 public class PrintException extends RuntimeException {
+    public enum StatusCode {
+        CoverOpened,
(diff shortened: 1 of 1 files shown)


openpos-peripheral/src/main/java/org/jumpmind/pos/print/PrintException.java
```diff
@@ -1,28 +1,73 @@
...
+@EqualsAndHashCode(callSuper = true)
 public class PrintException extends RuntimeException {
+    public enum StatusCode {
+        CoverOpened,
+        OutOfPaper
+    }
+
+    final EnumSet<StatusCode> status;
+
+    private static String createMessageFromStatus(EnumSet<StatusCode> status) {
```
jrmihalick, line 18: Is this beneficial to have the status message here?  I like the addition of the enum, but thought maybe that would be sufficient enough to allow callers to use i18n to look up locale-based message?

eddielegg-le: This is jut a cherry pick from what was in master, and there's some stuff in commerce that uses this logic

jrmihalick: ok, thanks for clarification.

jrmihalick (approved):

PR #1237, opened 2021-04-05
Improved Create and Edit Loyalty Customer dialogs layout, added full address and phone/email list input support

brandon-lewis:
### Summary
This change includes a remodeling of the Create and Edit Loyalty Customer dialogs from being generic "DynamicForm" components. The layout of the dialog form has been updated, and additional fields have been added (Address Line 1, Line 2, Address, City, State, Zip).

This new specific form component supports two new CTX_FORM_FIELD types: PHONE_LIST and EMAIL_LIST. These lists allow creating/editing a customer's phone and email addresses as a collective whole, as well as the type/label for each email/phone.

### Screenshots
Updated layout for the Create Loyalty Customer dialog. This screenshot shows off the new Email and Phone lists, as well as the address field additions.
CreateLoyaltyCustomer

The Edit Loyalty Customer dialog form showing what the Phone and Email lists look like when the user only has 1 phone and 1 email.
EditLoyaltyCustomer

The Edit Loyalty Customer dialog form displaying the single value Phone/Email fields instead of the new list types.
EditLoyaltyCustomer_Single


openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/loyalty-customer-form-dialog/loyalty-customer-form.component.ts
```diff
@@ -0,0 +1,125 @@
...
+import {IActionItem} from "../../core/actions/action-item.interface";
+import {Membership} from "../../shared/screen-parts/membership-display/memebership-display.interface";
+
+@Component({
+    selector: 'app-loyalty-customer-form',
+    templateUrl: './loyalty-customer-form.component.html',
+    styleUrls: ['./loyalty-customer-form.component.scss']
+})
+export class LoyaltyCustomerForm extends DynamicFormPartComponent {
+
```
dankaste, line 16: I would prefer that we not pass all these values in through inputs. It looks like for the most part the values coming into these inputs are coming in through the screen object and by using a screen part you can access them through the this.screenData object that the ScreenPartComponent abstract class creates for you.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/loyalty-customer-form-dialog/loyalty-customer-form.component.ts
```diff
@@ -0,0 +1,125 @@
...
+import {IFormElement} from "../../core/interfaces/form-field.interface";
+import {IActionItem} from "../../core/actions/action-item.interface";
+import {Membership} from "../../shared/screen-parts/membership-display/memebership-display.interface";
+
+@Component({
+    selector: 'app-loyalty-customer-form',
+    templateUrl: './loyalty-customer-form.component.html',
+    styleUrls: ['./loyalty-customer-form.component.scss']
+})
+export class LoyaltyCustomerForm extends DynamicFormPartComponent {
```
dankaste, line 15: Maybe I'm missing it but I'm not sure what you are gaining by extending DynamicFormPartComponent here. What that component does is gather all the FormElements from a form object on the screen and create the generic form. In your case I think you would be better off just extending ScreenPartComponnt and getting all your form elements from screenData. Then you can grab each element by name and bind it to the appropriate dynamic-form-field in the view.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/loyalty-customer-form-dialog/loyalty-customer-form.component.ts
```diff
@@ -0,0 +1,125 @@
...
+    memberships: Membership[] = [];
+    @Input()
+    membershipsLabel: string;
+    @Input()
+    noMembershipsLabel: string;
+
+    @Input()
+    isMobile: Observable<boolean>;
+
+    firstNameField : IFormElement;
```
dankaste, line 57: We should be using the Reactive Form pattern for forms where we use the FormBuilder Service to build our form create your dynamic-form-field elements give them the appropriate field name and wrap them all in an app-form. You then bind to the formSubmit event on app-form and this makes sure all the validation works correctly. You can then use formBuilder.buildFormPayload to sent the results back to the server.

dankaste (changes requested):
I think it would be better if the form usage here was more consistent with the rest of our forms. The only difference would be that instead of using dynamic-form-part which just lays all the elements out in rows you would have a custom one so you can control the layout. But we should still be using app-form, dynamic-form-field, formBuilder and payload should still a Form on the server as well. Theres lots of validations problems we've already solved in that code and I don't want to bring in a new pattern to maintain.

brandon-lewis:
@dankaste The pattern you suggested doesn't _entirely_ seem to fit in this situation given we're providing extra add/remove actions for the dynamic Phones and Emails lists. On top of this, we're presenting non-interactive Groups/Membership information. I want to adhere to the precedent/pattern you've identified here, so I have two questions:

1. For additional actions that aren't directly tied to a single FormElement, would it be appropriate to add a new  `Map<String, ActionItem>` map to Form.java (and a respective `formActions: IActionItem[];` to the IForm typescript interface) to allow passing in these extra actions to the form?
2. Since Memberships are non-interactive and not plain old "form elements", does it make sense to continue providing Membership-related data as component Inputs?

dankaste:
> @dankaste The pattern you suggested doesn't _entirely_ seem to fit in this situation given we're providing extra add/remove actions for the dynamic Phones and Emails lists. On top of this, we're presenting non-interactive Groups/Membership information. I want to adhere to the precedent/pattern you've identified here, so I have two questions:
> 
> 1. For additional actions that aren't directly tied to a single FormElement (but that are still necessary for the form to function), would it be appropriate to add a new  `Map<String, ActionItem>` map to Form.java (and a respective `formActions: IActionItem[];` to the IForm typescript interface) to allow passing in these extra actions to the form?
> 2. Since Memberships are non-interactive and not plain old "form elements", does it make sense to continue providing Membership-related data as component Inputs?

For 1. If I'm understanding correctly what you are trying to do. I wouldn't add the additional actions to the form since they are not part of the form but acting on the form. I would expect those on the screen/screen part along with the form.
For 2. If your component is a screen part you would just get it from screenData.

If you make a component a screen part you are giving it access to the screen view model that you are sending from the server and you can design it 1 of 2 ways. Either you get just a part of the screen as defined by the name you set in the decorator or you can actually just give it the whole screen and pick the interface you pass in as the generic parameter will "expose" just the properties on the screen you care about.

Perhaps a quick design session call might help.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/loyalty-customer-form-dialog/loyalty-customer-form.component.ts
brandon-lewis: As discussed, I've removed the extension of DynamicFormPart and moved away from the ScreenPartComponent entirely, moving the whole form itself to the PosScreen dialog component.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/loyalty-customer-form-dialog/loyalty-customer-form.component.ts
brandon-lewis: As mentioned in another comment, I've moved away from using the ScreenPartComponent and condensed the form to be on the dialog PosScreen component entirely.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/loyalty-customer-form-dialog/loyalty-customer-form.component.ts
brandon-lewis: The dialog PosScreen component now uses the app-form pattern mentioned here.

jrmihalick (approved):

dankaste (approved):

PR #1241, opened 2021-04-06
Allow properties to be removed from the ClientContext (support for fixing sequence number duplication) - cherry pick from releases/1.9

jrmihalick:
Cherry pick of changes from #1178 (releases/1.9)

### Issues Fixed
PDPOS-4786

### Summary
Openpos supporting changes for providing a solution for preventing duplicate sequence numbers during service calls between commerce instances.

- Turn down logging on service calls.
- Allow properties to be removed from the ClientContext


openpos-service/src/main/java/org/jumpmind/pos/service/EndpointInvoker.java
```diff
@@ -277,7 +277,7 @@ private void log(Method method, Object[] args, String implementation) {
                     }
                 }
             }
-            log.info("{}.{}() {}",
+            log.debug("{}.{}() {}",
```
mmichalek, line 280: Wonder if we should wrap this with log.isDebugEnabled() for effeciency

jrmihalick: @mmichalek Yeah, agree

mmichalek (approved):

PR #1244, opened 2021-04-08
Better logging for jackson serialization failures.  Exception gets swallowed.

chenson42:
### Summary
Better logging for jackson serialization failures.  Exception gets swallowed.

Was getting: Could not extract response. no suitable HttpMessageConverter found
Now we see the actual serialization error.


jrmihalick (approved):

stevencarley (approved):

chenson42:
This code came from this stackoverflow article.  https://stackoverflow.com/questions/44176335/restclientexception-could-not-extract-response-no-suitable-httpmessageconverte

@mmichalek I know it might overlap with some of the stuff you are doing.   Just making you aware.

PR #1245, opened 2021-04-08
Use barcode instead of receipt number when removing a receipt 

b-laci:


chenson42 (approved):

openpos-flow/src/main/java/org/jumpmind/pos/core/ui/data/TransactionReceipt.java
```diff
@@ -22,6 +22,16 @@
 
     private String icon;
 
+    private String barcode;
+
+    public String getBarcode() {
```
chenson42, line 27: While we are in here it would be nice to switch this to use lombok

PR #1252, opened 2021-04-09
Migrate Caching of Method Invocations From Master

JoeBrownlee:
### Issues Fixed
Support caching of method invocations in default and Training Modes. This is based on the existing functionality on master.

### Summary
Add the internal tracking of methods. As methods are encountered, they are cached to make future invocations more efficient. Maps are maintained for both default mode and Training Mode and the appropriate one is used based on the current device mode.

### DIFF
--- a/openpos-service/src/main/java/org/jumpmind/pos/service/strategy/LocalOnlyStrategy.java
+++ b/openpos-service/src/main/java/org/jumpmind/pos/service/strategy/LocalOnlyStrategy.java
@@ -1,17 +1,22 @@
 package org.jumpmind.pos.service.strategy;
 
 import org.jumpmind.pos.service.InjectionContext;
+import org.jumpmind.pos.util.clientcontext.ClientContext;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.jumpmind.pos.service.PosServerException;
-import org.jumpmind.pos.service.ServiceSpecificConfig;
 import org.springframework.stereotype.Component;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


openpos-service/src/main/java/org/jumpmind/pos/service/strategy/LocalOnlyStrategy.java
```diff
@@ -39,4 +52,12 @@ public Object invoke(List<String> profileIds, Object proxy, Method method, Map<S
                 + "with a method annotated like  @Endpoint(path=\"%s\")", path, path));
     }
 
+    private Map<Method, Method> getMethodMapForDeviceMode()  {
+        return (getDeviceMode().equals("training") ? trainingMethodMap : defaultMethodMap);
```
jrmihalick, line 56: You can use `TrainingModeHelper.isTraningMode(ClientContext)` for this.  It's in 1.9 and 1.10, but I haven't gotten it in master yet I think.

jrmihalick (changes requested):

openpos-service/src/main/java/org/jumpmind/pos/service/strategy/LocalOnlyStrategy.java
jrmihalick: Whoops disregard.  This is in openpos.  I should probably move TrainingModeHelper to Openpos.

jrmihalick (approved):

openpos-service/src/main/java/org/jumpmind/pos/service/strategy/LocalOnlyStrategy.java
JoeBrownlee: I agree, it would be way better to get TrainingModeHelper where pretty much everything can see it.

PR #1254, opened 2021-04-09
Cherry Pick Training Mode Changes From 1.9/1.10 To Master

JoeBrownlee:
### Issues Fixed
This feature is intended for use by all JumpMind customers. It addresses the following Petco JIRA items in whole or in part: #PDPOS-2759, #PDPOS-2823, #PDPOS-2824, #PDPOS-4117, #PDPOS-4241, #PDPOS-4243.

### Summary
This adds a greatly enhanced version of Training Mode to the OpenPOS Framework and Commerce. Users can train on most every function of the POS using real or "shadowed" data and services. This is highly configurable.

The basis of this change is a new concept of a **Device Mode**. The Device Mode for the POS is normally `default`. When the POS is in Training Mode, it is `training`. Other modes could be added in the future.

When in Training Mode, the POS can use **Shadow Tables** and **Shadow Endpoints**. Shadow tables are parallel versions of Commerce tables that are separate from the main production database tables. They have the same name and structure as the "normal" tables, except that they are prefixed with a special shadow prefix, `TNG_` by default. The prefix is configurable. The shadow tables are used while the POS is in Training Mode.

Likewise, shadow endpoints can be created to "override" normal POS behavior while in Training Mode. A shadow endpoint will have the same service path as the default endpoint, but its implementation will be `training`.

The `application.yml` file in the commerce-server project controls the configuration of shadow tables. Its `deviceMode` tag must match the Training Mode value `training`. It specifies the shadow table prefix, whether queries should be validated at startup, and tables to be included or excluded from shadowing. `*` can be used as a wildcard in table names. For example:
```
shadowTables:
    deviceMode: training
    tablePrefix: tng
    validateQueries: true
    includes:
      - sls_*
      - cust_*
      - ops_*
    excludes:
      - sls_rtn_policy
      - sls_tender
      - cust_phone_blacklist
  ```

For the POS to adjust SQL to use the shadow tables, the `*-query.yml` and `*-dml.yml` files need to have their SQL written to use the model class name for a given entity rather than a hard-coded table name. For example, a query like `select * from SLS_TRANS` would be written instead as `select * from TransModel`. As part of this work, all existing SQL in the query files has been changed to conform to this. Note _**one important exception**_: if a query has a join or sub select that uses a table _outside_ its own module, those table names must be left as the actual table name rather than the model class name. For example, in `sls-query.yml`, there is SQL that selects data from the `USR_USER` table. That reference must stay as the actual table name.

The GUI has been changed as well. In previous versions of Commerce, clicking the **Training Mode** button on the Home screen would allow the operator to perform one transaction and then they would exit Training Mode. Now, when the operator clicks the **Training Mode** button, the POS will be in Training Mode until the operator explicitly exits by returning to the Home screen and clicking the **Training Mode** button again. Entering Training Mode requires a manager's authorization. While in Training Mode, the color scheme changes to a grey palette and a watermark appears at the top of each screen to indicate the POS is in Training Mode.

This is the initial version of this feature. Future PRs will add additional functionality.

### Database Changes
There are no database changes per se, other than the fact that shadow tables will be created automatically at startup as specified in the `application.yml` file. Existing tables are not changed.

### CODE REVIEW

**Comment by jrmihalick** [comment]
Looks like at least some of the unit tests are failing due to this very helpful error  😄  :
```
    Caused by: org.***.db.sql.InvalidSqlException: SQL SELECT contains literal table name CAR_CAR in query carsByLiteralReplacementInSelectClause in file car-query.yml. Replace it with the corresponding model class name CarModel instead.
```

File: openpos-service/src/main/java/org/jumpmind/pos/service/EndpointInvoker.java [hunk 1]
```diff
@@ -280,7 +342,7 @@ private void log(Method method, Object[] args, String implementation) {
             log.info("{}.{}() {}",
```
jrmihalick, line 342: Let's make this a debug log message.  It makes the logs super lengthy with it being at info.  I already changed to debug on 1.9/1.10 previously.

jrmihalick (changes requested):

jrmihalick:
You can run all of the openpos unit tests locally using a command like this:

From the openpos-assemble directory:

```
> ./gradlew -PskipClient test
```

See my comment on your commerce PR for running the unit and cuc tests in commerce.

JoeBrownlee:
> Looks like at least some of the unit tests are failing due to this very helpful error 😄 :
> 
> ```
>     Caused by: org.***.db.sql.InvalidSqlException: SQL SELECT contains literal table name CAR_CAR in query carsByLiteralReplacementInSelectClause in file car-query.yml. Replace it with the corresponding model class name CarModel instead.
> ```

The CAR_CAR queries were on my work list, but somehow got lost. But yes, one of the changes is to prevent startup if there is a hard-coded table name -- assuming the `validateQueries` parameter is **true**.

openpos-service/src/main/java/org/jumpmind/pos/service/EndpointInvoker.java
JoeBrownlee: Done.

jrmihalick (approved):

PR #1259, opened 2021-04-12
[ISR] Swatch Image Fallback

kdelorey:
### Summary
Updated swatch picker to fallback to a drop-down picker if any of the swatch images are unavailable. 

image

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/item-detail/option-components/swatch-product-option-part/swatch-product-option-part.component.html
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/item-detail/option-components/swatch-product-option-part/swatch-product-option-part.component.html
@@ -1,15 +1,24 @@
 <h3>{{screenData.optionName}}</h3>
-<div class="swatch-container">
-    <div class="swatch" *ngFor="let swatch of screenData.swatches" 
-         (click)="selectOption(swatch)" 
-         [class.disabled]="swatch.disabled"
-         [class.selected]="swatch.id==screenData.selectedOption">
-        <div class="swatch-wrapper">
-            <img class="swatch-img"
-                 [src]="swatch.imageUrl">
+<div *ngIf="useImageSwatch; else imageSwatchBackup">
+    <div class="swatch-container">
+        <div class="swatch" *ngFor="let swatch of screenData.swatches" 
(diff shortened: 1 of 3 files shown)


openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/item-detail/option-components/swatch-product-option-part/swatch-product-option-part.component.html
```diff
@@ -1,15 +1,24 @@
...
         </div>
     </div>
-
-
+    <caption>{{selectedOptionName}}</caption>
 </div>
-<caption>{{selectedOptionName}}</caption>
\ No newline at end of file
+<ng-template #imageSwatchBackup>
+    <mat-select (selectionChange)="selectOption($event.value)" [value]="screenData.selectedOption">
```
dankaste, line 18: Should probably have a place holder here like we do for the basic-product-option-part when there is no selection

`<mat-select (selectionChange)="optionSelected($event)" [placeholder]="screenData.optionPlaceholder" [value]="screenData.selectedOption">
    <mat-option *ngFor="let option of screenData.options" 
                [value]="option.id" 
                >{{option.name}}</mat-option>
</mat-select>`

dankaste (changes requested):
Just add the place holder and we are good. Not sure why I didn't think to do it this way.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/item-detail/option-components/swatch-product-option-part/swatch-product-option-part.component.html
kdelorey: Good catch, I'll get it added.

kdelorey: image

kdelorey: Dependent on JumpMind/commerce#3559 to function correctly.

dankaste (approved):

PR #1263, opened 2021-04-13
Should always return implementation match for endpoint overrides

chenson42:
### Summary
Should always return implementation match for endpoint overrides

### DIFF
--- a/openpos-service/src/main/java/org/jumpmind/pos/service/EndpointInvoker.java
+++ b/openpos-service/src/main/java/org/jumpmind/pos/service/EndpointInvoker.java
@@ -194,23 +194,14 @@ public class EndpointInvoker implements InvocationHandler {
 
     protected Object findBestEndpointOverrideMatch(String path, String implementation, Collection<Object> endpointOverrides) {
         Object bestMatch = null;
-        List<SimpleEntry<Object,EndpointOverride>> pathMatchedOverrides = endpointOverrides.stream()
-                .map(o -> new SimpleEntry<>(o, ClassUtils.resolveAnnotation(EndpointOverride.class, o)))
+        List<SimpleEntry<Object, EndpointOverride>> pathMatchedOverrides = endpointOverrides.stream()
+                .map(o -> new SimpleEntry<Object, EndpointOverride>(o, ClassUtils.resolveAnnotation(EndpointOverride.class, o)))
                 .filter(entry -> entry.getValue().path().equals(path)
                 ).collect(Collectors.toList());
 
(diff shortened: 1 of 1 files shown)


openpos-service/src/main/java/org/jumpmind/pos/service/EndpointInvoker.java
```diff
@@ -339,7 +330,7 @@ private void log(Method method, Object[] args, String implementation) {
                     }
                 }
             }
-            log.debug("{}.{}() {}",
+            log.info("{}.{}() {}",
```
jrmihalick, line 333: I asked for this to be turned to debug because it ends up making a lot of output in logs.  At least on 1.9 and 1.10 it does.

jrmihalick (commented):
This seems to conflict with #1181, #1185.  1185 is still pending, but #1181 ended up coming over with the training mode merge.  Seems we still would have the problem that #1181 addressed.

openpos-service/src/main/java/org/jumpmind/pos/service/EndpointInvoker.java
chenson42: I put it at info on purpose.  It is useful and doesn't make the log that messy.

chenson42:
> This seems to conflict with #1181, #1185. 1185 is still pending, but #1181 ended up coming over with the training mode merge. Seems we still would have the problem that #1181 addressed.

Not sure what any of those are, but functionally you would never want to return the default implementation of an endpoint override if the implementation is set to something else ... unless you are in training mode.   I didn't address that here.

openpos-service/src/main/java/org/jumpmind/pos/service/EndpointInvoker.java
jrmihalick: Petco's logs were full of invocations to `IDevicesService.getMyDevice()` and some other endpoints.

jrmihalick:
To summarize #1181, the reason for the change in that PR was that if we had an implementation specified in the yaml for a service (e.g., returnsmgmt with implementation 'appriss'), endpoint overrides in petco were not getting run and the base commerce implementation was getting run instead.  I don't think the change you removed would cause a problem for training mode (because training mode endpoints are not overrides, they are a different implementation), but we will have to fix the issue identified by #1181.

chenson42:
I fixed the issue with with the endpoint override i was having trouble with in the demo project by deleting it.  I'll just close this PR as I don't think we should be using endpoint overrides anyways.   I'll just stay away from them.

PR #1267, opened 2021-04-14
Display loyalty reward amount values

r-cochran:
### Summary
On the customer details loyalty rewards and loyalty rewards history tabs, added the ability to show the aggregate value of transaction amount based loyalty rewards.

### Screenshots
image
image
image
image

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/rewards-history-line-item/_rewards-history-line-item-theme.scss
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/rewards-history-line-item/_rewards-history-line-item-theme.scss
@@ -4,4 +4,12 @@
     .reward-history-line-item-wrapper .status, .mobile-reward-history-line-item-wrapper .status {
         color: $primary;
     }
+
+    .reward {
+        &.expired {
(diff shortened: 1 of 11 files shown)


jrmihalick:
Did they want to see two UI text element to indicate that a reward is expired?  Both in a separate column and as dimmed text below the description?  Just wondering why we need both?

jrmihalick:
Ok, I see the dimmed line below the description is more of a status line.  Sometimes says "Expires m/d/yyyy" and then says "Expired m/d/yyyy" when expired.  Still debatable I think whether we really need the third column, but I understand now why we have the dimmed text.

r-cochran:
> Ok, I see the dimmed line below the description is more of a status line. Sometimes says "Expires m/d/yyyy" and then says "Expired m/d/yyyy" when expired. Still debatable I think whether we really need the third column, but I understand now why we have the dimmed text.

This PR involves just the second column of data, introducing the reward value. 

image

Was the image we were used to give direction for the reward history tab when we originally built it out. Since that tab shows both the used and expired rewards, I think the expiration date as a sub-line for the name is to give context as to when redeemed rewards would have expired (though I'm not sure how useful that is).

jrmihalick:
Ok, thanks for the context.

jrmihalick (approved):

PR #1272, opened 2021-04-16
Fix So Shadow Tables Are Created With Tagged/Augmented Columns

JoeBrownlee:
### Issues Fixed

Add support for shadow tables to create tag and/or augmented columns when they are created.

### Summary
When shadow tables were being created, they did not have any tag or augmented columns. When a table was created, it was only adding the columns for the default version of the table, not the shadowed one. This fix simply looks to see if there is a corresponding shadow table and if so, it does the same exact thing there that it does for the default one.

### Database Changes
None per se other than creating a shadow table with tag or augmented columns correctly.


jrmihalick (approved):
Please also wait for @chenson42 or @mmichalek review also.  Thx.

mmichalek (approved):

PR #1275, opened 2021-04-16
We were creating the connection pool for each module more than once

chenson42:
### Summary
We were creating the connection pool for each module more than once.  Also fixed logging in a new DbSession method.  The original exception needed to be included in the stack trace.


openpos-service/src/main/java/org/jumpmind/pos/service/AbstractRDBMSModule.java
```diff
@@ -248,7 +250,7 @@ protected static void loadJumpMindDriver() {
     @Override
     public DataSource getDataSource() {
```
stevencarley, line 251: Is this ever getting executed by multiple threads? If so think about making this synchronized.

stevencarley (approved):

PR #1278, opened 2021-04-19
Removed logo from the receipt

b-laci:


mmichalek (changes requested):
Why is this a "bugfix" in our base product?

chenson42:
@b-laci  This should be configurable.   Not flat out removed.   The icon should still show in the base product.

PR #1279, opened 2021-04-19
Feature/issue 3512

TyShingler:
### Issues Fixed
Commerce #3512

### Summary
Changes the sampling strategy from of using `@Sample` annotations to using the database entries in `CTX_CONFIG` and yaml configurations. Removes `@Sample` annotation from all endpoints and updates docs. Adds logic to several config classes for checking the data table `CTX_CONFIG` for overrides to that instance as well as adding class variables for sampling configurations where needed.

Currently the feature "Confirm you can add a config via database on a new endpoint" will not work!

### DIFF
--- a/openpos-service/src/main/java/org/jumpmind/pos/service/EndpointInvoker.java
+++ b/openpos-service/src/main/java/org/jumpmind/pos/service/EndpointInvoker.java
@@ -76,6 +76,7 @@ public class EndpointInvoker implements InvocationHandler {
             build();
 
     private static final ExecutorService instrumentationExecutor = Executors.newSingleThreadExecutor(factory);
+    protected HashMap<String, Boolean> endpointEnabledCache = new HashMap<String, Boolean>();
 
     @EventListener
     public void onApplicationEvent(ContextRefreshedEvent event) {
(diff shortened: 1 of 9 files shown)


openpos-service/src/main/java/org/jumpmind/pos/service/EndpointInvoker.java
```diff
@@ -363,22 +361,30 @@ private ServiceSpecificConfig getSpecificConfig(Method method) {
...
-        ServiceSampleModel sample = null;
-        if (method.isAnnotationPresent(Sample.class)) {
-            sample = new ServiceSampleModel();
-            sample.setSampleId(installationId + System.currentTimeMillis());
-            sample.setInstallationId(installationId);
-            sample.setHostname(AppUtils.getHostName());
-            sample.setServiceName(method.getDeclaringClass().getSimpleName() + "." + method.getName());
-            sample.setServiceType(strategy.getStrategyName());
-            sample.setStartTime(new Date());
+        Optional<ServiceSampleModel> serviceSampleModel = Optional.empty();
```
jrmihalick, line 370: I'm thinking we should keep a simple hashmap cache of the endpoint paths which have sampling enabled so we don't have to do this filtering constantly, do you agree @chenson42 and @mmichalek ?

mmichalek: Sure, caching would be great.

jrmihalick (changes requested):
See comment above

openpos-service/src/main/java/org/jumpmind/pos/service/EndpointInvoker.java
TyShingler: @jrmihalick We talked about the cacheing, would you prefer using a hashmap over say the spring annotation, cacheable? It is used for cashing in ServiceConfig.java on getServiceConfig().

jrmihalick: You can use spring @Cacheable if you want.  Just can't be in the EndpointInvoker class due to spring proxying.

TyShingler: Using cacheable seems like I would have to add unnecessary complexity then. Will use the hashmap solution.

jrmihalick (approved):

jrmihalick (approved):

PR #1285, opened 2021-04-20
Device landscape styles not being applied

blubaugha:
### Summary
Added logging to help troubleshoot landscape mode CSS classes not being applied at Burlington on their POS device.


JoeBrownlee (approved):
Looks good to me, but I'll let Jason be the final authority.

jrmihalick (approved):

PR #1321, opened 2021-05-06
Current class can be null.  Protect against it (#1205)

mmichalek:
### Summary
Cherry pick: current class can be null. protect against it (#1205)


jrmihalick (approved):

jchirico8:
This sounds like the title of a talk at a tech conference

mmichalek:
That's all Chris :)

PR #1328, opened 2021-05-07
Use correct Device import for device info

eddielegg-le:
### Summary
Incorrect import for Device plugin used to get device name. Corrected import.


jrmihalick:
Let's hold on merging this for just a bit until Petco actually gets 1.9.4 release deployed.  It's held up at the moment by an alleged promotions problem.

kdelorey (approved):

PR #1344, opened 2021-05-12
make inventory calls async

dankaste:
### Summary
Leverage the data providers to send the inventory data async to the client as to not hold up the sending of the initial screen.


chenson42 (approved):

kdelorey (dismissed):
I think this needs to be merged to `release/2.1`. I think that have some changes in `release/2.1` that will cause some conflicts, so watch for that.

kdelorey (approved):

PR #1348, opened 2021-05-14
PCOM-3155 add property field for notify message (need to implement an…

yuying6688:
PCOM-3155 add property field for notify message

This is for adding notify message display on item detail page. Since my client dev local env not work, the Angular part in detail html will be done later.


kdelorey (dismissed):

chenson42 (dismissed):

kdelorey (approved):

dankaste (commented):
Shouldn't this be going into releases/2.2

kdelorey:
Don't forget there's a merge conflict here.

PR #1351, opened 2021-05-14
Fix a bug that made it so we couldn't load templates from multiple paths

dankaste:
### Summary
We had an issue where we couldn't load templates from multiple paths, this fixes it.


chenson42 (approved):

JoeBrownlee (commented):
Makes sense, if there are already temples in that mode, add them. Otherwise load them in.

PR #1354, opened 2021-05-17
Added buttons to flip forwards and backward across the last 128 receipts

levshunovm:
### Summary
Added buttons to flip forwards and backward across the last 128 receipts. Buttons are disabled if there isn't next or previous receipt respectively.

### Screenshots
Screenshot 2021-05-18 at 16 29 43
Screenshot 2021-05-18 at 16 30 26


chenson42:
@levshunovm instead of having two receipts I think it would be better to keep a list of the preview tokens in the screen and add buttons to flip forwards and backwards across receipts.  It looks like the html is kept around for 128 previous receipts.

levshunovm:
@chenson42 re-implemented following your suggestion. I tried to find how to access previous values kept by the HTML as you said but found only suggestions to store it yourself, so I've done it and use 128 as a maximum number of stored receipts. Let me know If you think it should be a different number.

chenson42 (approved):

dankaste (approved):

PR #1365, opened 2021-05-24
Added repository logic for beans functionality

apomanytskyi:
Added repository logic for beans functionality.

How it works example:

```
class TypeCode {
    A, 
    B
}
```

```

public interface SomeSpecificForSomeEnumTypeLogicService {
    TypeCode getCode();
    void executeLogic();
}
```

```
@Component
public class AService implements SomeSpecificForSomeEnumTypeLogicService {
 
    @Override
    public TypeCode getCode() {
        return TypeCode.A;
    }
    @Override
    public void executeLogic() {
         System.out.println("Some logic specific for TypeCode.A type");
    }
}
```
```
@Component
public class BService implements SomeSpecificForSomeEnumTypeLogicService {
 
    @Override
    public TypeCode getCode() {
        return TypeCode.B;
    }
    @Override
    public void executeLogic() {
         System.out.println("Some logic specific for TypeCode.B type");
    }
}
```

So we can create repository like this:

```
@Component
public class TypeCodeRepository extends ServiceRepositoryImpl<TypeCode, SomeSpecificForSomeEnumTypeLogicService> {

    @Autowired
    protected void setServices(List<SomeSpecificForSomeEnumTypeLogicService> services) {
        super.setServices(services, SomeSpecificForSomeEnumTypeLogicService::getCode);
    }
}
```

And in some our class we can execute specific logic related with specific TypeCode

```
@Autowired 
TypeCodeRepository repository;

public void someMethod(TypeCode typeCode) {
    repository.getService(typeCode).executeLogic(); 
    // If typeCode param = TypeCode.A - will fire:  AService.executeLogic
    // if typeCode param = TypeCode.B - will fire:  BService.executeLogic
}
```


mmichalek (commented):
Could this pluggable logic be accomplished by referring to a service with a bean name?

apomanytskyi:
> Could this pluggable logic be accomplished by referring to a service with a bean name?

@mmichalek The idea was to be able to get beans directly by constants/enums link. 
Expressions cannot be used in component names, so instead of a direct reference to a constant/enum - you have to use its text value.
If the constant changes, then the functional by name breaks if it is not corrected.
It was just a suggestion. I reworked my logic through the names of the beans - so I decline from this commit.

PR #1367, opened 2021-05-24
Loyalty Button layout changes

r-cochran:
### Summary
Changed the layout/styling to the loyalty button when a customer is linked.

### Screenshots
Default
image
image

dark-blue (the icon being white is an artifact of the image for the inverted logo colors being white, this would be tailored per customer)
image
image

pets
image
image

training
image
image

patriotic bird
image
image

patriotic nest
image
image

burl red
image
image

bang orange
image
image

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/membership-display/_membership-display-theme.scss
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/membership-display/_membership-display-theme.scss
@@ -4,15 +4,16 @@
     $openpos-membership-in-background: map-get($theme, openpos-membership-in-background );
 
     .membership-chips {
-        mat-chip.not-in {
-            background-color: white !important;
-            border: 1px solid mat-color($foreground, secondary-text);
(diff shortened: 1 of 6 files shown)


jrmihalick (approved):

r-cochran:
> Looks good, Petco wanted the shortened name of the groups to be displayed on the Loyalty button. Will that be possible with this implementation?

good callout I had forgotten that part. I believe it was discussed as using the ID instead of the name of the group. We currently don't have a short name field for the customer groups table, but might be something to expand into.

mmichalek:
I don't think we can use group id for this. I think we will need a "shortDescription" field on the group model.  If it's populated, then we can use it there on the sale screen.  If it's not populated, then use the current displayValue.

jrmihalick:
Now is the time to add it, I think.  QA will probably flag it anyway.

r-cochran:
made the changes, but they were all commerce side. @mmichalek was there anything else to tweak on the openpos side?

mmichalek (approved):

mmichalek:
Looks good Robert.

PR #1374, opened 2021-05-26
One Document at a Time Printing

kdelorey:
### Summary
This is part of a change to resolve an issue where printing on iPads would cause the multiple end-of-day print jobs to sort of cancel each other out because there was no print queueing. Now the EOD printing will only ask for one document at a time to be printed. When it's done printing an action to signal the to the server that it's done printing and will cause the server to send the next document for printing.

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/printers/browser-printer.plugin.ts
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/printers/browser-printer.plugin.ts
@@ -26,7 +26,6 @@ export class BrowserPrinterPlugin implements IPrinter {
                 };
 
                 contentWindow.onafterprint = () => {
-                    console.log('after print.');
                     observer.next();
                     observer.complete();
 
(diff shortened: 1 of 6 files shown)


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/printers/browser-printer.plugin.ts
```diff
@@ -13,26 +14,38 @@ export class BrowserPrinterPlugin implements IPrinter {
...
 
-        hiddenPrintFrame.style.position = "fixed";
-        hiddenPrintFrame.style.right = "0";
-        hiddenPrintFrame.style.bottom = "0";
-        hiddenPrintFrame.style.width = "0";
-        hiddenPrintFrame.style.height = "0";
-        hiddenPrintFrame.style.border = "0";
-        hiddenPrintFrame.srcdoc = html;
+                contentWindow.onafterprint = () => {
+                    console.log('after print.');
```
jchirico8, line 29: Do you still need this log? It's not adding anything

kdelorey: Oh no, good catch. I was debugging.

kdelorey: fixed.

jchirico8 (dismissed):

kdelorey:
Rebased onto `releases/2.1`

chenson42 (approved):

PR #1375, opened 2021-05-27
Make remove button not be cut off on return linked receipt

levshunovm:
### Issues Fixed
At some devices remove button cut off on the return linked receipt:
image-2021-05-04-08-42-20-862

### Summary
I noticed that the mat-icon-button is hardcoded to be 40px/40px which is not enough for some devices in our case. Also, I found out that we have a bin icon close to the remove button which was cut off on all devices, so my fix also make it visible.

### Screenshots
Comparasion with and without changed width:
Screenshot 2021-05-19 at 21 47 16
Screenshot 2021-05-19 at 21 47 37

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/receipt-card/receipt-card.component.html
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/receipt-card/receipt-card.component.html
@@ -46,7 +46,7 @@
     <div class="remove-button" *ngIf="receipt && removeReceiptAction">
         <button mat-icon-button (click)="onRemoveAction(); $event.stopPropagation()" color="primary">
             {{removeReceiptAction.title}}
-            <app-icon [iconClass]="'material-icons'" [iconName]="removeReceiptAction.icon"></app-icon>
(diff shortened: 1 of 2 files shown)


chenson42:
@blubaugha @jchirico8 Not sure if hard coding the size is the correct answer here?  Can you guys make a suggestion?

jchirico8:
This seems to happen because `.mat-icon-button.md` is in _icons.scss as 40px. lg is 64px and xl is 128px. My opinion is those default sizes for a button are far too small and they seem to be more of a standard for the icon itself, not the whole button

levshunovm:
@jchirico8, so you suggest using some different class here instead?

blubaugha:
How about trying something like setting the `min-width` to the content? That way when the language and/or text change the button won't cut off the text.

Something like this might work:

image

You might need to fix the icon to remove the extra white space at the top, or put in a little offset tweak. I had to shift the icon up a few pixels (see screenshot) to make it vertically align with the text.

levshunovm:
@blubaugha, thanks for the suggestion! Implemented it, please review.

chenson42 (approved):

PR #1380, opened 2021-06-01
Enforce the use of API tokens for REST API endpoints

mzipay:
### Summary
Invoking an exposed REST API over http(s) now requires an authorization token conveyed in the "X-API-Token" header.  Exclusions are supported based on Ant-style path patterns configured in _application.yml_.  Additionally, all stock REST APIs that were previously mounted under "/path" are now mounted under "/rest/path" (a permanent redirect will be issued for requests that use the old paths).

### Parameter Changes
The "X-API-Token" header is now a required parameter for all exposed REST APIs.

### Screenshots
image


mzipay:
Please see also:
https://github.com/JumpMind/commerce/pull/3925

chenson42 (approved):

PR #1382, opened 2021-06-07
Updated Customer Details Rewards tab display and added Reward keyboard navigation

brandon-lewis:
### Summary
Added keyboard navigation to the Customer Details Rewards tab (up and down to select a row, Spacebar to apply the selected reward). Also added a general "applied/status" icon column to provide extra reward-related context if so desired.

### Screenshots
The screenshot below shows what the keyboard navigation now looks like when navigating through available rewards. The list auto-selects/highlights the first reward available. Selecting the reward with the Spacebar applies the reward and selects either the next available reward (or the previous if none exists).

Applied rewards are disabled as before but now show a different icon. The new status icon to the left of the reward amount is an optional configurable detail provided by State controllers.

CustDetails_AppliedHighlights

When no rewards are available to be applied to the transaction, the list stops being navigable, the buttons remain disabled as before, and the reward selection indicator disappears.

CustDetails_NoMoreAvailable


openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/UILoyaltyReward.java
```diff
@@ -11,10 +11,12 @@
 public class UILoyaltyReward implements Serializable {
     private static final long serialVersionUID = 1L;
 
+    private boolean isLoadedReward;
```
jrmihalick, line 14: Could we just have a `statusIcon` column instead of a boolean.  And if that statusIcon is set to a value it displays the icon?  Let me know if that doesn't make sense?  I want to get away from the Loaded terminology, since, as far as we know at this point, it is a petco-specific concept.

jrmihalick (changes requested):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/sale/customer-details-dialog/customer-details-dialog.component.ts
```diff
@@ -32,9 +55,62 @@ export class CustomerDetailsDialogComponent extends PosScreen<CustomerDetailsDia
...
+      reward.enabled = (reward.applyButton && reward.applyButton.enabled == true);
+      this.allRewards.set(i, reward);
+      if(!reward.enabled) {
+        this.allDisabledRewards.set(i, reward);
+      }
+    }
+
+      this.listData = new Observable<ISelectableListData<Reward>>((observer) => {
+        observer.next({
+          items: this.allRewards,
```
kdelorey, line 86: Capturing `this` and reusing the same collection was a cleaver solution to your problem. That being said I still think the ergonomics of `app-selectable-item-list` could be improved by allowing it respond to changes of the `[listData]` correctly. I actually don't think it should take an `Observable<>` at all as in input and instead take the list of data directly. Using it with an `Observable<>` would simply mean adding a `async` pipe to the input. Anyways, we can punt on this, I just wanted to bring it up.

kdelorey (commented):
LGTM, I'm going to let @jrmihalick approve this since I saw he had a comment that was resolved.

jrmihalick (approved):

PR #1388, opened 2021-06-10
Add endpoint that returns all devices

lutzb-edje:


openpos-devices/src/test/java/org/jumpmind/pos/devices/service/GetAllDevicesEndpointTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Test
+    public void getDeviceShouldReturnMatchingDevice() throws Exception {
+
+        String result =
+            mvc.perform(new MockGetRequestBuilder("/devices/").build())
+            .andExpect(status().isOk())
+            .andReturn().getResponse().getContentAsString();
+
+        assertEquals(10, mapper.readValue(result, GetAllDevicesResponse.class).getDevices().size());
```
jrmihalick, line 43: Should this perhaps make an independent call to the DevicesRespository to getAllDevices and establish the expected count of devices that way instead of hard-coding a value of 10?  I am assuming 10 is based off our current test data that we are loading up.

jrmihalick (changes requested):

openpos-devices/src/test/java/org/jumpmind/pos/devices/service/GetAllDevicesEndpointTest.java
lutzb-edje: Good call. I'll fix it

jrmihalick (approved):

PR #1389, opened 2021-06-10
Add static hostname fallback to auto-personalization

eddielegg-le:
### Summary
Add a fallback to the auto-personalization to check for a statically configured hostname and endpoint for the auto-personalization config. If the call fails or the config is null, will fall back to manual personalization.

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/auto-personalization-startup-task.ts
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/auto-personalization-startup-task.ts
@@ -9,6 +9,7 @@ import {PersonalizationComponent} from '../personalization/personalization.compo
 import {Zeroconf, ZeroconfService} from "@ionic-native/zeroconf";
 import {StartupTaskNames} from "./startup-task-names";
 import {WrapperService} from "../services/wrapper.service";
+import {Configuration} from "../../configuration/configuration";
 
 
 @Injectable({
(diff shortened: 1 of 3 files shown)


maxwellpettit (approved):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/auto-personalization-startup-task.ts
```diff
@@ -26,30 +27,51 @@ export class AutoPersonalizationStartupTask implements IStartupTask {
...
                     }),
                     flatMap(() => this.wrapperService.getDeviceName()),
                     tap(deviceName => name = deviceName),
-                    flatMap(() => this.attemptAutoPersonalize(serviceConfig, name)),
-                    catchError(() => this.manualPersonalization())
+                    flatMap(() => {
+                        const url = `${serviceConfig.hostname}:${serviceConfig.port}/${serviceConfig.txtRecord.path}`
+                        return this.attemptAutoPersonalize(url, name)
+                    }),
+                    catchError(() => this.personalizeWithHostname())
```
jrmihalick, line 50: We don't seem to log errors or any messages when we fail at each place.  Let's add some more logging so we can figure out why we fail when we do.

jrmihalick (changes requested):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/auto-personalization-startup-task.ts
eddielegg-le: I think the more detailed logging is going to be on the server side for the most part. I'll see if there's some good client side logging opportunities

jrmihalick (approved):

PR #1391, opened 2021-06-11
Fallback to specified hostname for auto-personalization

eddielegg-le:
### Summary
Add a fallback to the auto-personalization to check for a statically configured hostname and endpoint for the auto-personalization config. If the call fails or the config is null, will fall back to manual personalization.


jrmihalick (approved):

stojankz (approved):

kdelorey (approved):
I tested this against AEO's stuff, as far as I can tell it seems to be working.

PR #1393, opened 2021-06-11
Connecting to Server... dialog only closed after screen message comes through 

kdelorey:
### Summary
This resolves a bug that was discovered where it was possible for a user to execute an action between the time the client connected to a server and the server sent a screen message (~5s in between). The screen messages are already trying to close the loading dialog. This change just defers the time at which the connection itself cancels the dialog.


kdelorey:
@chenson42 can you verify what branch this is supposed to go to? I have it on `releases/2.1` but I'm not sure what branch is supposed to get this where you want it.

chenson42 (approved):

chenson42:
> @chenson42 can you verify what branch this is supposed to go to? I have it on `releases/2.1` but I'm not sure what branch is supposed to get this where you want it.

Branch is correct

PR #1396, opened 2021-06-14
Allow only first 14 digits for money

apomanytskyi:
Allow only first 14 digits for money.

Before:
item_price_before

After:
item_price_after


apomanytskyi:
In PR https://github.com/JumpMind/commerce/pull/3915 we already have price configuration, so, in this PR - it's just UI restriction to not have values like: 1.23+22E

chenson42 (approved):

PR #1415, opened 2021-06-19
Fix tender and personalization screen bugs

yuying6688:
Fix the following UI bugs:

1. Tender screen positions, responsive on the a proper position:

image

2. Fix personalization screen dropdown transparent, added mouse hover item by highlight:

image

image

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/personalization/personalization.component.scss
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/personalization/personalization.component.scss
@@ -1,0 +1,15 @@
+.dropdown{
+    background-color: #faf9f9;
+}
+
+::ng-deep{
+    .mat-option-text:hover:not(.mat-option-disabled) {
+        background: lightgray !important;
+        color: gray;
+    }
+
+    .mat-option:hover {
+        background: lightgray !important;
+        color: gray !important;
+    }
+}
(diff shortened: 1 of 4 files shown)


blubaugha (dismissed):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/personalization/personalization.component.scss
```diff
@@ -0,0 +1,15 @@
+.dropdown{
+    background-color: #faf9f9;
+}
+
+::ng-deep{
+    .mat-option-text:hover:not(.mat-option-disabled) {
+        background: lightgray !important;
```
jchirico8, line 7: Do we need all of these importants? This selector seems pretty specific that it shouldn't overlap with another style

yuying6688: Right, removed the third !important which is no use. But still need second !important du ethe mat-option and mat-option-text are not in same box.

dankaste (dismissed):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/personalization/personalization.component.scss
jchirico8: I'm specifically referring to the `!important` part. I don't mind having the background color defined in multiple places, but it's bad practice to use `!important` because it forces the style onto elements you might not realize could be affected

yuying6688: Fixed

jchirico8 (approved):

PR #1416, opened 2021-06-21
Updated reward-line-item component to support status text instead of an icon

brandon-lewis:
### Summary
Updated the RewardsLineItem components (part of CustomerDetails) to take reward-based action item details instead of screen level, and updated the optional status icon to be text instead.

### Screenshots
A sample reward that can have a customized status text displayed (in the example below, this is a "Fancy Reward").
image

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/sale/customer-details-dialog/customer-details-dialog.component.scss
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/sale/customer-details-dialog/customer-details-dialog.component.scss
@@ -76,6 +76,7 @@ app-content-card {
   .membership-icon {
     .material-icons{
       margin-top: -10px;
+      width: 80px;
     }
   }
   .memberships {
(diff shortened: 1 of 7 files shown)


stojankz:
Might not be related to your changes, but could you take a look at the spacing that's happening here?
Screen Shot 2021-06-21 at 11 45 34 AM

brandon-lewis:
> Might not be related to your changes, but could you take a look at the spacing that's happening here?
> Screen Shot 2021-06-21 at 11 45 34 AM

@stojankz This is a bit of an unlucky screenshot for this page and a combination of CSS settings. First, those grid sections are spaced with fractional grid spacing. Second, the profile icon is only 80x80 svg in a 90x90 container, whereas the memberships icon is a 90x67.5 svg in a 90x90 container, so it looks tighter because that memberships icon is bigger than the profile icon. This styling has been in since the end of March.

stojankz:
> > Might not be related to your changes, but could you take a look at the spacing that's happening here?
> > Screen Shot 2021-06-21 at 11 45 34 AM
> 
> @stojankz This is a bit of an unlucky screenshot for this page and a combination of CSS settings. First, those grid sections are spaced with fractional grid spacing. Second, the profile icon is only 80x80 svg in a 90x90 container, whereas the memberships icon is a 90x67.5 svg in a 90x90 container, so it looks tighter because that memberships icon is bigger than the profile icon. This styling has been in since the end of March.

Right I figured it wasn't a part of your current change, but if we could add a little bit of padding or so between the icon and text I think it would be a lot better visually. Figured it would be a small change but if not, don't worry about it

stojankz (approved):

stojankz (approved):

stojankz (approved):

PR #1427, opened 2021-06-24
Customer Item History Filtering

kdelorey:
### Summary
Added filtering capabilities to customer item history.


jchirico8:
Is it too late to keep your stuff in feature/real-rewards? AEO threw a fit when I put part of the rewards stuff in the base branch so they're making us put everything into that feature for now

kdelorey:
> Is it too late to keep your stuff in feature/real-rewards? AEO threw a fit when I put part of the rewards stuff in the base branch so they're making us put everything into that feature for now

Probably, I already have most of it in `2.2` & `develop`. This was just something we tacked on and I put it into a different PR. :(

I could have it turned off (via. configuration) and then in the `feature/real-rewards` branch we turn it on?

dankaste (approved):

PR #1428, opened 2021-06-25
Item lookup - made selecting a color change the main image

levshunovm:
### Summary
This change is needed to make setter in the carousel.component.ts detect changes. It checks only a memory link for an array that is not changing by default. There's a way to override that check itself but it affects performance badly. Let me know if you think it's still better to override the check or if you have a better solution.

Before change:
before

After change:
after

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/item-detail/item-detail.component.ts
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/item-detail/item-detail.component.ts
@@ -71,6 +71,7 @@ export class ItemDetailComponent extends PosScreen<ItemDetailInterface> {
             .pipe(map( stores => stores != null && stores != undefined));
         this.inventoryMessage$ = this.dataMessageService.getData$(this.screen.inventoryMessageProviderKey)
             .pipe(map(value => value != null ? value[0] : null));
+        this.screen.imageUrls = [].concat(this.screen.imageUrls);
     }
     
     getComponentFromOptionType(productOption: ProductOptionInterface) {


chenson42 (approved):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/item-detail/item-detail.component.ts
```diff
@@ -71,6 +71,7 @@ export class ItemDetailComponent extends PosScreen<ItemDetailInterface> {
             .pipe(map( stores => stores != null && stores != undefined));
         this.inventoryMessage$ = this.dataMessageService.getData$(this.screen.inventoryMessageProviderKey)
             .pipe(map(value => value != null ? value[0] : null));
+        this.screen.imageUrls = [].concat(this.screen.imageUrls);
```
dankaste, line 74: I'm not super crazy about this... no one in the future is going to know why you are doing this.
I'm looking into this real quick. There are ways to manually trigger change detection when we call the setter. I was pretty sure that the ngFor would be looking at the elements in the array and those should be changing.

dankaste (changes requested):

dankaste (approved):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/item-detail/item-detail.component.ts
dankaste: Never mind... you are right here. I see no better way with how it is currently designed.

PR #1467, opened 2021-07-13
Fix overflow issue on suspended txn list

eddielegg-le:
### Issues Fixed
https://petcoalm.atlassian.net/browse/PDPOS-5983

### Summary
Fixed issue with transaction list overflow. Dialog was pushing the select button off page and was causing a double scroll issue inside the dialog content. Changing max-height resolved both issues. Also removed unnecessary div.

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/selection-list/selection-list-screen-dialog.component.html
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/selection-list/selection-list-screen-dialog.component.html
@@ -4,43 +4,41 @@
         <app-instructions *ngIf="screen.instructions" [instructions]="screen.instructions" [instructionsSize]="'text-sm'"></app-instructions>
         <app-scan-or-search *ngIf="screen.showScan" class="scan-or-search mat-elevation-z1"></app-scan-or-search>
         <app-content-card>
-            <mat-dialog-content>
-                <div class="dialog-content">
-                    <app-selectable-item-list  *ngIf="screen.selectionList && screen.selectionList.length > 0" fxFlex="0 1 100%" [listData]="listData"
-                        [configuration]="listConfig" [defaultSelect]="screen.defaultSelect"
-                        [selectedItemList]="selectedItems"
-                        [selectedItem]="selectedItem"
-                        (selectedItemChange)="onItemChange($event)" (selectedItemListChange)="onItemListChange($event)"
-                        class="item-list">
-                        <ng-template let-item>
-                            <div #items gdAreas="img info" gdColumns="auto 2fr" [ngClass]="{'muted-color' : !item.enabled}">
-                                <div gdArea="img">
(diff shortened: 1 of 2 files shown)


stojankz:
I wonder if that max height should adjust based on media type? Ie how does it look if you switch from desktop to iPad to mobile?

eddielegg-le:
> I wonder if that max height should adjust based on media type? Ie how does it look if you switch from desktop to iPad to mobile?

Suppose it could - I could use the ngStyle breakpoints to change the max-height property

stojankz:
> > I wonder if that max height should adjust based on media type? Ie how does it look if you switch from desktop to iPad to mobile?
> 
> 
> 
> Suppose it could - I could use the ngStyle breakpoints to change the max-height property

We actually have a way to do this in the scss, if you search other components you should be able to see examples

eddielegg-le:
> > > I wonder if that max height should adjust based on media type? Ie how does it look if you switch from desktop to iPad to mobile?
> > 
> > 
> > Suppose it could - I could use the ngStyle breakpoints to change the max-height property
> 
> We actually have a way to do this in the scss, if you search other components you should be able to see examples

Right, just seemed more straight forward to use the breakpoints and we're also using that. Is the scss the preferred way?

stojankz:
> Right, just seemed more straight forward to use the breakpoints and we're also using that. Is the scss the preferred way?

I guess personally I think it's cleaner to have it in the scss so the html doesn't get to cluttered, but if it works then it's probably okay how you have it

stojankz:
Here's an example from transaction-summary.component.scss:

```
    &.tablet-landscape,
    &.tablet-portrait,
    &.mobile-landscape,
    &.mobile-portrait,
    &.desktop-portrait {
        width: 48px;
        height: 48px;
    }
```

eddielegg-le:
> Here's an example from transaction-summary.component.scss:
> 
> ```
>     &.tablet-landscape,
>     &.tablet-portrait,
>     &.mobile-landscape,
>     &.mobile-portrait,
>     &.desktop-portrait {
>         width: 48px;
>         height: 48px;
>     }
> ```

Appreciate it, was more of a question of preference since we are currently using both methods.

I looked into it a bit further and neither approaches were working completely. It looks like this may have been a situation where the screen size was non-standard and the height may have hit the media breakpoint for a tablet but the width did not, so the max-height never changed. I ended up using more specific media selectors for screen height to change the max-height instead.

stojankz (approved):

stojankz:
> > Here's an example from transaction-summary.component.scss:
> > ```
> >     &.tablet-landscape,
> >     &.tablet-portrait,
> >     &.mobile-landscape,
> >     &.mobile-portrait,
> >     &.desktop-portrait {
> >         width: 48px;
> >         height: 48px;
> >     }
> > ```
> 
> Appreciate it, was more of a question of preference since we are currently using both methods.
> 
> I looked into it a bit further and neither approaches were working completely. It looks like this may have been a situation where the screen size was non-standard and the height may have hit the media breakpoint for a tablet but the width did not, so the max-height never changed. I ended up using more specific media selectors for screen height to change the max-height instead.

looks good, thanks for looking into it some more Eddie

PR #1481, opened 2021-07-20
Internal refactor to allow patching of query.yml files

mmichalek:
### Issues Fixed
This makes it possible to patch query yml files in a patch.jar. Also unified loading logic a little between the dml & query yamls.


openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSessionFactory.java
```diff
@@ -149,32 +144,51 @@ public static QueryTemplates getQueryTemplates(String tablePrefix) {
...
+            Enumeration<URL> urls = Thread.currentThread().getContextClassLoader().getResources(resourceName);
+
+            if (!urls.hasMoreElements()) {
+                log.debug("Could not locate " + resourceName + " on the classpath.");
+                return new ArrayList<>();
             }
+
+            List<URL> resources = Collections.list(urls);
+            // reverse the order from classpath so that higher ordered files can applied last and thus take effect.
+            // MMM 07-20-2021: Additional note, there probably isn't a strong reason the query yml's couldn't fit
```
jrmihalick, line 186: This was a confusing comment to me at first, but I think you are just saying we could have instead use spring resource loading for the query yml files instead of our own loading.

jrmihalick (approved):

stojankz (approved):

openpos-persist/src/main/java/org/jumpmind/pos/persist/DBSessionFactory.java
mmichalek: Yeah I was just thinking out loud that we've used application*.yml successfully for a lot now, and probably could have fit the queries into that structure as well.

PR #1491, opened 2021-07-27
Feature/api tokens and rest path

mzipay:
### Summary
Support for API token auth and "/rest" context path (rebased onto 1.12).


openpos-util/src/main/java/org/jumpmind/pos/util/RestApiSupport.java
```diff
@@ -0,0 +1,145 @@
...
+        String legacyRequestMapping = restApiRequestMapping.substring(REST_API_CONTEXT_PATH.length());
+        int contextSlash = legacyRequestMapping.indexOf('/', 1);
+        return contextSlash == -1 ? legacyRequestMapping : legacyRequestMapping.substring(0, contextSlash);
+    }
+
+    private static Class<?> classForName(final String className) {
+        try {
+            return Class.forName(className);
+        } catch (ClassNotFoundException ex) {
+            LOG.warn(ex.toString());
```
mmichalek, line 137: Let's change this to something like
` LOG.warn("Failed to find class " + className, ex); so we get the whole stack trace.`

openpos-util/src/main/java/org/jumpmind/pos/util/RestApiSupport.java
```diff
@@ -0,0 +1,145 @@
...
+    public static final String REST_API_TOKEN_HEADER_NAME = "X-API-Token";
+
+    /**
+     * @see org.jumpmind.pos.util.ApiTokenGenerator#generateApiToken()
+     */
+    public static final String REST_API_LOOPBACK_TOKEN = "LjVHdqj47ecbHVnkBB35";
+
+    private static final Function<String, String> REQUEST_MAPPING_TO_WILDCARD_URI_PATTERN = (requestMapping) -> requestMapping + "/*";
+
+    private static final Logger LOG = LoggerFactory.getLogger(RestApiSupport.class);
```
mmichalek, line 33: Can replace with @Slf4j annotation on the class, and use "log" variable.

openpos-util/src/main/java/org/jumpmind/pos/util/RestApiSupport.java
```diff
@@ -0,0 +1,145 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.Collections;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public final class RestApiSupport {
```
mmichalek, line 21: Can this be a Spring bean vs. static singleton?

openpos-util/src/main/java/org/jumpmind/pos/util/RestApiSupport.java
mzipay: An entire stack trace is a bit much in this case.  Contextually, the log entry as-is will already note where the incident occurred (RestApiSupport), what went wrong (ClassNotFoundException) and the FQCN of the class that wasn't found. A full stack trace here will not provide anything additionally meaningful (and may actually be quite confusing to a log viewer as this is simply a WARN-level record, not an error).
Are you sure you want a full stack trace here?

openpos-util/src/main/java/org/jumpmind/pos/util/RestApiSupport.java
mzipay: It isn't a singleton - there is no instance, just a group of interrelated (static) utility methods that are used in lambdas and one factory method. It does not have state; not really a good candidate for a bean.

openpos-util/src/main/java/org/jumpmind/pos/util/RestApiSupport.java
mzipay: Also note that this is used in the context of a collection stream - one of the reasons it's kept short and sweet.

openpos-util/src/main/java/org/jumpmind/pos/util/RestApiSupport.java
mzipay: Done.

openpos-util/src/main/java/org/jumpmind/pos/util/RestApiSupport.java
mmichalek: Just use ClassUtils.loadClass() instead please.

openpos-util/src/main/java/org/jumpmind/pos/util/RestApiSupport.java
mmichalek: We cannot easily plug in replacement code for static methods, whereas we can override object-based beans.  If this can work as a bean, please make it a bean.

mmichalek (changes requested):

openpos-util/src/main/java/org/jumpmind/pos/util/RestApiSupport.java
mzipay: I think we should discuss this further, because I don't quite follow why this would be a bean (or need to be).  The package org.jumpmind.pos.util is full of classes just like this one, doing the same thing (collecting related stateless static utility methods). What is the reasoning behind turning _this_ utility class into a bean? Perhaps I don't understand the motivation or some overarching goal.

Worth noting there are already mechanisms here for changing behavior in the spring beans that _use_ this utility. For example see the overloads for requestMappingToUrlPattern and also the fact that the injectable ApiTokenAutoConfiguration#restApiScanner(String) bean simply delegates to RestApiSupport#newRestApiScanner - which is trivial to modify post-construction or even to completely swap out the entire impl.

Can you take a look at the `@Bean` methods in ApiTokenAutoConfiguration? That's where the "replacement" of any of this utility logic would occur. I think that may satisfy what (I think) you're looking for here.

openpos-util/src/main/java/org/jumpmind/pos/util/RestApiSupport.java
mzipay: Done.

mmichalek (approved):

jrmihalick:
Had a conversation with Matt about these endpoints which are called directly from angular client code:

**ping**: openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/session.service.ts
**fileUpload**: openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/file-upload.service.ts
**fetchMessage**: openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/fetch-message.service.ts
**personalize**: openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/personalization/personalization.service.ts

personalize and ping are the most important to be excluded from the token api.  fileUpload is only used by one retailer on an old version of openpos.  fetchMessage was written by @dankaste , so I am not sure how it is used.

jrmihalick (approved):

PR #1496, opened 2021-07-28
Audio Preload Fixes

JJStarrett:
Previously, the audio service sometimes had issues with startup. This was often one of two main issues - 

 - In some cases (nearly always when the service is first started on a fresh browser cache), the list of audio files for preload sent over to the client during startup would not all be successfully loaded into the browser's cache. The client `AudioService` would refuse to finish startup, however, until all files were preloaded successfully. This often meant the service would effectively never start. This was my most encountered issue.
 
 - Due to the way ORPOS's startup happens in combination with ours, the `ContentProviderService` may not have been created when the client's `AudioService` sends its startup config and preload requests. In this case, we get a NPE and the client `AudioService` can not start or starts with the default config - which has audio disabled by default. 

So, I've made the following adjustments:

- The client now times out after 25 seconds of preloading. If it can not load all audio files in 25 seconds, it will timeout and continue starting up. Preloading is not necessary, _functionally_, to audio (we can still play audio requests even if they were not preloaded) and it has been adjusted on the client side to behave as such.

- When the server's `AudioService` receives a request for a list of audio files to preload / the audio config during startup, it will now try to wait until the `ContentProviderService` is created before attempting to fill those requests. The thread processing the startup message will retry every second for up to 45 seconds to get the `ContentProviderService`. As far as I can tell, this runs concurrently to other tasks and does not adversely affect other startup tasks as audio is the last to run, seemingly to try to avoid this issue (priority of 950, and `FinalStartup` has a priority of 1000).


jrmihalick (approved):
Can't help but wonder if this should have been implemented differently originally so as to avoid the race condition which you are fixing.

JJStarrett:
> Can't help but wonder if this should have been implemented differently originally so as to avoid the race condition which you are fixing.

I kind of agree. Startup here seemed a little optimistic. Preloading is nice when it's successful, but it's a bit unreliable and probably shouldn't have been blocking the rest of the service's startup. I'm also not totally sure why the server _and_ the client need a copy of the audio config at startup. Or why the client override's the server's config. The server could probably make decisions alone, I would think. 

Nevertheless, should be a bit more robust now and hopefully all goes to smoothly for Burlington.

PR #1503, opened 2021-07-30
Feature/audio preload server error handling

JJStarrett:
### Summary

When the client requests a list of files to preload from the server, the server attempts to auto-populate a list of suspected audio keys. It does this by looking at the folder structure the audio files are stored in. While a noble cause, this can often hit some errors when executed inside of a jar file, so error handling and additional attempts at auto-populating the keys have been added.

Ultimately, preload is not vital to audio functionality. We're catching all exceptions here because if preloading fails, we don't want the whole service to go down.


openpos-flow/src/main/java/org/jumpmind/pos/core/audio/AudioUtil.java
```diff
@@ -61,40 +68,76 @@ public static String getKey(String sound) {
...
+        } catch (Exception error) {
+            e = error;
+        }
+        if (StringUtils.isEmpty(key)) {
+            try {
+                URL url = resource.getURL();
+                String path = url.getPath();
+                key = path.substring(path.indexOf(AUDIO_CONTENT_ROOT), path.indexOf(resource.getFilename()));
+                key = key.replace(AUDIO_CONTENT_ROOT, "");
+            } catch (Exception error) {
```
jrmihalick, line 122: Looks like `e` could have been set above on line 114 (to `error`), but not cleared here and then end up throwing an exception on line 127 even though key could have been set on line 121?

jrmihalick (commented):
Do you know why are we getting errors when reading from the jar file?  Or do you need these changes to determine that?

JJStarrett:
> Do you know why are we getting errors when reading from the jar file? Or do you need these changes to determine that?

We're searching a file path to try to find the audio keys automatically. The keys are the parent folders of the resources. So when we called Paths.get() in a jar file, we errored out because, as far as I can tell, the resource being pathed to is technically not a file when in a jar. The URI ends up being something like: `jar:file:/C:/path/to/jar/file.jar!/audio/please-scan-voice/en_US/file.mp3` which Paths.java can't seem to parse.

But, I think we can try to deduce the folder without trying to convert the URI to a file path. I think for the future it would probably make sense to adjust how we get keys. We should probably keep a list somewhere or use the licenses.csv file to specify configurations (like localization). We would probably need to solve some localization problems in that case as well. For now though, the service doesn't depend on preloading succeeding, so its not vital that we always get the keys correct.

openpos-flow/src/main/java/org/jumpmind/pos/core/audio/AudioUtil.java
JJStarrett: Oh you're right, we need to check that key is empty before we throw the exception. I've committed a change.

mmichalek:
I'm not sure I understand the background here, but 1 thing: 
`jar:file:/C:/path/to/jar/file.jar!/audio/please-scan-voice/en_US/file.mp3`
is NOT a file as you mentioned, but rather a classpath resource.
You can get the contents of it with code like this:
`Thread.currentThread().getContextClassLoader().getResource("").openStream()`

jrmihalick (approved):

JJStarrett:
> I'm not sure I understand the background here, but 1 thing:
> `jar:file:/C:/path/to/jar/file.jar!/audio/please-scan-voice/en_US/file.mp3`
> is NOT a file as you mentioned, but rather a classpath resource.
> You can get the contents of it with code like this:
> `Thread.currentThread().getContextClassLoader().getResource("jar:file:/C:/path/to/jar/file.jar!/audio/please-scan-voice/en_US/file.mp3").openStream()`

Right, but I think we're just looking for the folder the resource is contained in here. The `please-scan-voice/en_US`, which I think ends up being more of a string operation than anything. I think our problem was we were treating it as a file as a middle step to looking at the directory structure.

PR #1504, opened 2021-08-02
Add overridability to form field elements

JJStarrett:
### Issues Fixed
 - [5891](https://petcoalm.atlassian.net/browse/PDPOS-5891)

### Summary
Adds the ability to override a form field element that would normally be required. Overridability is controlled in `CTX_FORM_FIELD.OVERRIDABLE`. The `"Opt out"` label can be set on a field by field basis. All text input elements are eligible to be overridden. 

### Database Changes
- Added a column `Overridable` on `CTX_FORM_FIELD`

### Screenshots
a61f2132508f9c02ef7b2a6c79898d68


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/dynamic-form-field/dynamic-form-field.component.ts
```diff
@@ -237,6 +241,25 @@ export class DynamicFormFieldComponent implements OnInit, OnDestroy, AfterViewIn
     ).catch( error => console.info(`Scanning failed: ${error}`));
   }
 
+  isRequired(): boolean { 
+    return this.formField.required && !true;
```
r-cochran, line 245: Will this code always return false?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/dynamic-form-field/dynamic-form-field.component.ts
```diff
@@ -124,6 +124,10 @@ export class DynamicFormFieldComponent implements OnInit, OnDestroy, AfterViewIn
         this.formGroup.get(this.formField.id).setValue(this.formField.value);
       }
     }
+    if(this.formField.overridden) {
+      this.formField.overridden = false;
```
r-cochran, line 128: is the inversion of the overridden status intended or is the possibly just converting a truthy object into a literal false value?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/dynamic-form-field/dynamic-form-field.component.ts
stojankz: agreed, I'm guessing this should be something like 
```suggestion
    return this.formField.required && !this.formField. overridable;
```

JJStarrett: Ha, whoops. Yes, it will. Accidentally left some recent temp code in. This should have been `!this.formField.overridden`. Thanks for the catch, Robert, adjusted.

JJStarrett: Ha, whoops. Yes, it will. Accidentally left some recent temp code in. This should have been `!this.formField.overridden`. Thanks for the catch guys, adjusted.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/dynamic-form-field/dynamic-form-field.component.ts
JJStarrett: It's intended, though I think it was a bit sloppy. Previously, I was reusing the `onFormElementOverridden()` method which is normally called on checkbox check, and so would flip the `overridden` flag. 

Now, I've refactored the code that actually marks the field as not required / disabled into a separate helper method. The logic should be a bit less haphazard.

openpos-flow/src/main/java/org/jumpmind/pos/core/model/FormField.java
```diff
@@ -38,6 +38,9 @@
 
     private boolean preValidate = false;
 
+    private boolean isOverridable = false;
```
jchirico8, line 41: Would it be clearer to call this `isRequirednessOverridable` or something that explains it's meant for skipping required fields?

stojankz: Requiredness is not a word :P

JJStarrett: Well, I figured if you really wanted to, you _could_ override a non-required field. I'm not sure why you would, but it would be supported.

stojankz: ok so yeah to Joe's point, I do agree that isOverridable name is not clear in terms of _what_ is overridable. So maybe renaming it to isRequiredOverridable would be clearer

JJStarrett: Adjusted.

jchirico8 (approved):

PR #1544, opened 2021-08-20
Unit tests for Console Scraper

kdelorey:
### Summary
Adds unit tests for the console scrapper and a few assurances about the arguments.


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/logging/console-scraper.service.ts
```diff
@@ -48,14 +48,16 @@ export class ConsoleScraper {
...
-        if (args.length <= 0) {
+        if (!args || args.length <= 0) {
             return;
         }
 
+        let message = '';
+
         args.forEach(arg => {
-            if (typeof arg == 'object') {
+            if (!arg) {
```
kdelorey, line 58: This will ignore the literal `false`, probably handle null and undefined explicitly.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/logging/console-scraper.service.ts
```diff
@@ -48,14 +48,16 @@ export class ConsoleScraper {
...
+                // ignore falsey arguments
+            } else if (typeof arg == 'object') {
```
stojankz, line 60: since that if isn't doing anything could this just be changed to 
```suggestion
       if (arg && typeof arg === 'object') {
```

stojankz (changes requested):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/logging/console-scraper.service.ts
kdelorey: I don't want this to fall into the `else` condition either.

stojankz: Okay then I'd suggest this rather than have the blank if: 
```suggestion
        args.forEach(arg => {
            if (arg !== null || arg !== undefined) {
               if (typeof arg === 'object') {
                  let stringify = this._jsonStringifySanitize(arg);

                  if (stringify.length > 0) {
                      if (message.length > 0) {
                          message += '; ';
                      }
                      message += stringify;
                  }
               } else {
                  if (message.length > 0) {
                      message += '; ';
                  }
                  message += arg;
               }
           }
        });
```

jchirico8 (approved):

stojankz (approved):

PR #1563, opened 2021-08-27
Extra form fields styling "fix"

jchirico8:
### Issues Fixed
[PCOM-4424](https://jira.ae.com/browse/PCOM-4424) - Rows are overlapping on "Add Returns Customer" window

### Summary
This was broken in https://github.com/JumpMind/openpos-framework/pull/1539 which was trying to fix the loyaltyAdd styling. I imagine my PR here will be quoted in some future style fix.

### Screenshots
image
image
image

PR #1567, opened 2021-08-30
Fix add customer fields alignment

VodianytskyiVI:
Fix add customer fields alignment.
________________________________
Before:
Before
________________________________
After:
After
________________________________
Add Loyalty customer form stays unchanged:
AddLoyaltyCustomer


jchirico8:
I already did this in #1563. I've just been waiting on reviews

PR #1570, opened 2021-09-01
Fix add customer form field alignment

VodianytskyiVI:
______________________________________
Before:
Before
______________________________________
After:
After
______________________________________
Using 'Tab' button:
TabButton
______________________________________
Add New Loyalty user form stays unchanged:
AddNewLoyaltyUserForm
______________________________________


jchirico8:
When you tab through these fields does it still go in the order of first name, last name, line 1, line 2...?

VodianytskyiVI:
> When you tab through these fields does it still go in the order of first name, last name, line 1, line 2...?

I changed the order of fields, now first column will contain first half of extra elements, left - second half
Please take a look at the updated description with gif

chenson42:
@blubaugha @dankaste @stojankz @jchirico8 Can you please review?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/loyalty-customer-form-dialog/loyalty-customer-form-dialog.component.html
```diff
@@ -61,8 +61,9 @@
                                 </div>
                             </div>
                             <div class="extraFormFields">
-                                <ng-container *ngFor="let formElement of screen.form.formElements">
-                                    <app-dynamic-form-field *ngIf="handledFormFields.indexOf(formElement.id) == -1"
+                                <ng-container *ngFor="let formElement of extraElements; let i = index">
```
jchirico8, line 64: Could you add a `let length = screen.form.formElements` instead of creating a separate duplicate array? I don't know for sure that it would work

jchirico8: Or does the handledFormFields cause problems with that

VodianytskyiVI: Yes handledFormFields causes problems with numerations when some of elements are handled  and some aren't (a.k.a. extra fields)

jchirico8 (approved):

PR #1605, opened 2021-09-17
Cast Exception occurred when change configure in property file

chadhav-ae:
### Issues Fixed
Class Cast exception will occurred when we put the printer width in application-env.properties file. Property file return it as a String.


chenson42:
It would be nice to put that logic in a place where it would not be duplicated, but it looks like it would take a bit of refactoring in order to accomplish that cleanly.

chenson42 (approved):

PR #1606, opened 2021-09-17
Action sync support

mmichalek:
### Issues Fixed
This PR modifies the framework to link actions with the current state of the state machine.  If the state machine gets out of sync, and an action from a previous screen/state is raised, than the state machine will log and throw away that action (and refresh the screen.)

This change required client side changes to reports, because reports were sending down the action name as the report name, and then assigning that report name to the action data.  This changes things so the actionName/data go down and come up the same way.


openpos-flow/src/main/java/org/jumpmind/pos/core/flow/ActionHandlerImpl.java
```diff
@@ -47,6 +54,26 @@ public boolean handleAction(IStateManager stateManager, Object state, Action act
         }
     }
 
+    public boolean checkActionSync(IStateManager stateManager, ActionContext actionContext) {
+        if (actionContext.getSyncState() == null
+                || actionContext.getSyncState().equals(GLOBAL_ACTION)
```
jrmihalick, line 59: Should this perhaps instead be a boolean on the actionContext, `isGlobalAction()` instead of overloading the purpose of the syncState?

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/ActionSyncMessagePropertyStrategy.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public final static String SYNC_STATE_DIVIDER = "@";
+    public final static String GLOBAL_ACTION = "Global";
+
+    @Override
+    public Object doStrategy(String appId, String deviceId, Object value, Class<?> clazz, UIMessage message, Map<String, Object> messageContext) {
+        if (value instanceof ActionItem) {
+            ActionItem actionItem = (ActionItem) value;
+            String stateLabel = getStateLabel(actionItem);
+            if (stateLabel != null && !actionItem.getAction().contains(SYNC_STATE_DIVIDER)) {
```
jrmihalick, line 25: This logic
`actionItem.getAction().contains(SYNC_STATE_DIVIDER)` is used in at least a couple of places, should it be centralized somewhere in a utility class or the like?

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/LoggingStateManagerObserver.java
```diff
@@ -52,33 +52,48 @@ public void onTransitionStep(ApplicationState applicationState, Transition trans
...
-            screenLogger.info("Show screen on device \"" + applicationState.getDeviceId() + "\" (" + screen.getClass().getName() + ")\n"
-                    + drawBox(screen.getId(), screen.getScreenType()));
+            screenLogger.info(new StringBuilder().append("Show screen on device \"").
+                    append(applicationState.getDeviceId()).append("\" (").
+                    append(screen.getClass().getName()).append(")").append(getScreenSequence(screen)).append("\n").
+                    append(drawBox(screen.getId(), screen.getScreenType())).toString());
+        }
+    }
+
+    private String getScreenSequence(UIMessage screen) {
```
jrmihalick, line 84: Mucho thanko. 👍

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/StateManager.java
```diff
@@ -730,13 +737,21 @@ private boolean validateAction(Action action) {
...
     boolean isOnStateManagerThread() {
         return Thread.currentThread().getName().startsWith("StateManager");
     }
 
     protected void processAction(ActionContext actionContext) {
         lastInteractionTime.set(new Date());
+
+        boolean syncOk = actionHandler.checkActionSync(this, actionContext);
+        if (!syncOk) {
+            return;
```
jrmihalick, line 750: Should something be logged here or will that happen downstream from here?

jrmihalick (changes requested):

openpos-flow/src/main/java/org/jumpmind/pos/core/ui/ActionItem.java
```diff
@@ -58,11 +57,16 @@
     protected String additionalIcon;
     protected String additionalText;
     protected ActionTimer actionTimer;
+    protected Object data;
```
jrmihalick, line 60: Wonder if we should use `defaultPayload` for this instead?  Tough call, but it's close to the same usage.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/actions/action-item.ts
```diff
@@ -10,6 +10,7 @@ export class ActionItem implements IActionItem {
     doNotBlockForResponse?: boolean;
     buttonSize?: string;
     fontSize?: string;
+    data?: string;
```
jrmihalick, line 13: See my comment below about `defaultPayload`.  It's on the `IActionItem` interface here.

mmichalek: Agreed, I will go with defaultPayload.

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/ActionHandlerImpl.java
mmichalek: So the "sync state" is supposed to be slightly more abstract than a specific state name.  Maybe there's a better name.  I was thinking the future I wanted to leave this open even for a little more data e.g. 
Next@StateName
Next@7804d7 // e.g. a hash pointing to more complex state data
Next@StateName:otherData

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/ActionSyncMessagePropertyStrategy.java
mmichalek: Agreed

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/LoggingStateManagerObserver.java
mmichalek: 👍

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/StateManager.java
mmichalek: Yes - the checkActionSync will log a WARN in this case.

openpos-flow/src/main/java/org/jumpmind/pos/core/ui/ActionItem.java
mmichalek: Ok, will switch to defaultPayload.

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/ActionHandlerImpl.java
jrmihalick: Ok, I see. `syncyThingy`

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/LoggingStateManagerObserver.java
```diff
@@ -52,33 +52,47 @@ public void onTransitionStep(ApplicationState applicationState, Transition trans
...
+        if (screen.getOptionalProperties() != null && screen.getOptionalProperties().containsKey("sequenceNumber")) {
+            return " seqNo: "+screen.getOptionalProperties().get("sequenceNumber");
+        } else {
+            return "seqNo: ?";
         }
     }
 
     public void onTransition(ApplicationState applicationState, Transition transition, ITransitionStep currentTransitionStep) {
         logTranistionStep(transition, currentTransitionStep);
```
jrmihalick, line 92: You didn't modify this code, but I just noticed the method has a typo in it.  Can fix some other time, just FYI in case you want to fix it now.

jrmihalick (dismissed):

jrmihalick (approved):

PR #1613, opened 2021-09-21
Add locationData to the ClientContext

jrmihalick:
### Issues Fixed
PDPOS-6452

### Summary
- Client location data was not available for retail calculations for the first item added to the transaction in Commerce. This is because the `TransMgmtClient` handles setting the `TransLocationModel` on the transaction, but not until after the transaction has already been created and has the first item added to it.
- This adds the `LocationData` onto the `ClientContext` as a json string so that endpoints have access to the `LocationData`.
- Moved the `LocationData` class to `openpos-util` so that consumers do not have to have a dependency on `openpos-flow`, where it was previously located.


r-cochran (approved):

openpos-util/src/main/java/org/jumpmind/pos/util/model/LocationData.java
```diff
@@ -50,6 +50,14 @@ public void setLongitude(String longitude) {
         this.longitude = longitude;
     }
 
+    public static LocationData fromString(String jsonLocationData) {
+        try {
+            return DefaultObjectMapper.defaultObjectMapper().readValue(jsonLocationData, LocationData.class);
```
mmichalek, line 55: Small concern here is new'ing up a an ObjectMapper on each request. I think the ObjectMapper could be a static field as it's promised to be thread safe.
https://stackoverflow.com/questions/3907929/should-i-declare-jacksons-objectmapper-as-a-static-field

jrmihalick: Should we change that in the DefaultObjectMapper class and make it a singleton there?

mmichalek: 🤯

mmichalek: can't argue with that logic

mmichalek: Actually, maybe better to leave as is so callers can tweak the default configuration to their needs without disturbing other usages of that default object mapper.

mmichalek (approved):

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/LocationChangedGlobalActionHandler.java
```diff
@@ -17,6 +17,8 @@
     public void onLocationChanged(Action action) {
         LocationData locationData = Action.convertActionData(action.getData(), LocationData.class);
         clientLocationService.setLocationData(locationData);
+
+        stateManager.getClientContext().put("locationData", locationData != null ? locationData.toString() : null);
```
chenson42, line 21: Aren't there more places than just here you need to set the ClientContext?

chenson42: Like in StateManagerContainer.setCurrentStateManager

jrmihalick: I don't think so?  StateManagerContainer reads that map and sticks the entries onto the ClientContext.  See `StateManagerContainer.setCurrentStateManager(IStateManager)`

chenson42: ah

chenson42 (approved):

PR #1630, opened 2021-09-28
Match sql text to table name regardless of case

eddielegg-le:
### Summary
Currently matching SQL text with table names as case sensitive. Missing table names in query templates if the case does not match the db instance format.


openpos-persist/src/main/java/org/jumpmind/pos/persist/impl/AbstractSqlTemplates.java
```diff
@@ -20,10 +20,11 @@ protected String replaceClassModelNameInSqlText(String sqlText, String modelClas
 
     protected void scanSqlTextForLiteralTableName(String sqlText, String textType, String templateName, String regularTableName, String shadowTableName, String modulePrefix, String modelClassName)  {
         if (sqlText != null) {
-            boolean textContainsRegularName = sqlText.matches(".*\\b" + regularTableName + "\\b.*");
+            String lowerCaseSqlText = sqlText.toLowerCase();
+            boolean textContainsRegularName = lowerCaseSqlText.matches(".*\\b" + regularTableName.toLowerCase() + "\\b.*");
```
jrmihalick, line 24: Should also be able to use `sqlText.matches("(?i).*\\b" ....` and avoid the lower casing calls.

jrmihalick (approved):

openpos-persist/src/main/java/org/jumpmind/pos/persist/impl/AbstractSqlTemplates.java
eddielegg-le: If it's all the same I'd rather not add more regex 😆

PR #1635, opened 2021-09-29
Disabled auto everything for iOS/android keyboards.

kdelorey:
### Summary
Disable auto completes on every input for mobile keyboards.


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/directives/disable-auto-complete.directive.ts
```diff
@@ -0,0 +1,15 @@
+
+import { Directive } from '@angular/core';
+
+@Directive({
+    selector: 'input',
```
jchirico8, line 5: Is this auto applying to all inputs in the app?

kdelorey: Yes.

kdelorey: (unless otherwise overriden)

PR #1689, opened 2021-10-22
Made Transaction Search mobile-friendly

brandon-lewis:
### Summary
Cherry-picked changes to update Transaction Search to be mobile-friendly.

Includes cherry-picks from 1.12 Pull Requests: https://github.com/JumpMind/openpos-framework/pull/1559, https://github.com/JumpMind/openpos-framework/pull/1575

### Screenshots
Transaction Search on mobile.
image

Transaction Search Electronic Journal on mobile.
image

Unchanged Transaction Search on desktop.
image

Unchanged Transaction Search Electronic Journal on desktop.
image


blubaugha (dismissed):
# a
# P
# p
# R
# o
# V
# e
# !

DAVID HASSELHOFF

kdelorey (dismissed):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/transaction-summary/transaction-summary.component.spec.ts
```diff
@@ -8,31 +8,37 @@ import { OpenposMediaService } from '../../../core/media/openpos-media.service';
...
+    }
+  }
+
   beforeEach(async () => {
     await TestBed.configureTestingModule({
       imports: [MatChipsModule, HttpClientTestingModule],
       declarations: [
         TransactionSummaryComponent,
         MockIconComponent,
+        MockIconButtonComponent,
```
jchirico8, line 35: I added ng-mocks in here. Could you try using MockComponent instead of creating whole fake components? It should be cleaner and work the same

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/transaction-summary/transaction-summary.component.spec.ts
```diff
@@ -8,31 +8,37 @@ import { OpenposMediaService } from '../../../core/media/openpos-media.service';
...
   let fixture: ComponentFixture<TransactionSummaryComponent>;
 
+  class MockOpenposMediaServiceMobileFalse {
+    observe(): Observable<boolean> {
+      return of(false);
+    }
+  }
+
   beforeEach(async () => {
     await TestBed.configureTestingModule({
```
jchirico8, line 30: Does this need to be async when the only thing happening in the function is an await?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/transaction-summary/transaction-summary.component.spec.ts
brandon-lewis: I hadn't seen ng-mocks before, that's very convenient. Running the tests with changes now, will push soon.

> I added ng-mocks in here. Could you try using MockComponent instead of creating whole fake components? It should be cleaner and work the same

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/transaction-summary/transaction-summary.component.spec.ts
kdelorey: the `configureTestingModule` is a `Future<>` and is `await`ed. That being said, it shouldn't be entirely necessary but I started having weird timing issues when I did the Angular 11 upgrade and I basically started adding that `async`/`await` pattern to all fixtures.

brandon-lewis: @kdelorey I just pushed changes removing the async/await piece, still passing and running okay. Though this spec isn't doing much at the moment. If the async/await is a precautionary pattern we're trying to do everywhere to prevent those timing issues, I don't mind adding it back in.

jchirico8 (approved):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/transaction-summary/transaction-summary.component.spec.ts
jchirico8: I think ideally we don't want it if we don't need it, so as long as this spec file is passing consistently we should leave it out

jchirico8 (approved):

PR #1534, opened 2021-08-17
Updates to enhance support for one server and multiple clients having different business unit IDs

jrmihalick:
### Summary
- Adds support for personalizing (either manually or automatically) with a businessUnitId so that multiple clients with differing business unit IDs can connect to a single server
- Removes direct injection of the `openpos.businessunitId` from _most_ (not all) classes and instead changes classes to rely on the `ClientContext` to get the `businessUnitId`
- Introduces `openpos.personalization.deviceBusinessUnitIdStrategy` configuration parameter.  See below under **Parameter Changes** for more information.

### Database Changes
For support of setting the `businessUnitId` on the `DeviceModel` during auto-personalization, a `business_unit_id` column has been added to `dev_device_personalization`.  The value from this table for a given device will be used when the `openpos.personalization.deviceBusinessUnitIdStrategy` is set to `GetBusinessUnitIdFromPersonalizationParamsStrategy`.

### Parameter Changes
`openpos.personalization.deviceBusinessUnitIdStrategy` has been added to allow for configuration of the algorithm used for setting the business unit id on the `DeviceModel` that is created during personalization.  The default strategy is `GetBusinessUnitIdFromConfigStrategy`, which uses the configuration value of `openpos.businessunitId`.  Other strategies provided in this PR are:
- `ParseBusinessUnitIdFromDeviceIdStrategy` - extracts the business unit id from the device id provided during personalization
- `GetBusinessUnitIdFromPersonalizationParamsStrategy` - gets the business unit id from personalization parameters provided during manual personalization or during auto personalization.  Note that this also requires specifying a personalization parameter for the businessUnitId as follows:
```
openpos:
  ui:
    personalization:
      parameters:
        - property: 'businessUnitId'
          label: 'Business Unit ID'
        # other personalization parameters may follow
...
```

PR #1711, opened 2021-11-01
Updates to enhance support for one server and multiple clients having different business unit IDs - master

jrmihalick:
Cherry pick of #1534 to master

### Summary
- Adds support for personalizing (either manually or automatically) with a businessUnitId so that multiple clients with differing business unit IDs can connect to a single server
- Removes direct injection of the `openpos.businessunitId` from _most_ (not all) classes and instead changes classes to rely on the `ClientContext` to get the `businessUnitId`
- Introduces `openpos.personalization.deviceBusinessUnitIdStrategy` configuration parameter.  See below under **Parameter Changes** for more information.

### Database Changes
For support of setting the `businessUnitId` on the `DeviceModel` during auto-personalization, a `business_unit_id` column has been added to `dev_device_personalization`.  The value from this table for a given device will be used when the `openpos.personalization.deviceBusinessUnitIdStrategy` is set to `GetBusinessUnitIdFromPersonalizationParamsStrategy`.

### Parameter Changes
`openpos.personalization.deviceBusinessUnitIdStrategy` has been added to allow for configuration of the algorithm used for setting the business unit id on the `DeviceModel` that is created during personalization.  The default strategy is `GetBusinessUnitIdFromConfigStrategy`, which uses the configuration value of `openpos.businessunitId`.  Other strategies provided in this PR are:
- `ParseBusinessUnitIdFromDeviceIdStrategy` - extracts the business unit id from the device id provided during personalization
- `GetBusinessUnitIdFromPersonalizationParamsStrategy` - gets the business unit id from personalization parameters provided during manual personalization or during auto personalization.  Note that this also requires specifying a personalization parameter for the businessUnitId as follows:
```
openpos:
  ui:
    personalization:
      parameters:
        - property: 'businessUnitId'
          label: 'Business Unit ID'
        # other personalization parameters may follow
...
```


openpos-util/src/main/java/org/jumpmind/pos/util/clientcontext/ClientContext.java
```diff
@@ -45,7 +47,7 @@ public String get(String name) {
         if (props == null || !props.containsKey(name)) {
             if ("deviceId".equalsIgnoreCase(name)) {
```
chenson42, line 48: should we be adding constants for everything else to be consistent?

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/StateManager.java
```diff
@@ -138,6 +138,9 @@
     @Autowired
     EventPublisher eventPublisher;
 
+    @Autowired
+    ClientContext deviceClientContext;
```
chenson42, line 142: should this be clientContext versus deviceClientContext?

jrmihalick: There's already a `clientContext` attribute in StateManager which is being used from SessionSubscribedListener to pass in client context variables obtained from the `sessionAuthTracker` (SessionConnectListener instance).  For example to pass the client `timeZoneOffset` through to the StateManager.  It's pretty confusing.

chenson42: would renaming stuff a bit make it less confusing?

openpos-util/src/main/java/org/jumpmind/pos/util/clientcontext/ClientContext.java
jrmihalick: Yeah, I would be in favor of that.  I think it would change a lot of classes, though, because we dip into the `ClientContext` in a lot of places to get variable values.  I'd rather do that in a separate PR.  The reason I did businessUnitId as a constant was because I have seen places where we screw up the case of the variables or make a typo in the name.  The backing map in ClientContext is case-insensitive, so it's really for typos in this case.

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/StateManager.java
jrmihalick: Yeah, I think we need another name for the pre-existing `clientContext` map in the StateManager.  Maybe something like `deviceVariables`, `deviceParameters` ???

chenson42 (approved):

PR #1717, opened 2021-11-03
Bugfix/filter precedence

gwilmer:
### Summary
Add precedence for http request filters.  The LegacyApiRedirectFilter must be run last.  The LegacyApiRedirectFilter must do a forward for 1.11 vs a redirect because the 1.11 code doesn't handle the redirect.  It appears the forward doesn't allow the other filters to run if the LegacyApiRedirectFilter runs first.  This change runs the LegacyApiRedirectFilter last.

### DIFF
--- a/openpos-util/src/main/java/org/jumpmind/pos/util/clientcontext/ClientContentExtractionFilter.java
+++ b/openpos-util/src/main/java/org/jumpmind/pos/util/clientcontext/ClientContentExtractionFilter.java
@@ -4,6 +4,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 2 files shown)


chenson42 (approved):

openpos-util/src/main/java/org/jumpmind/pos/util/clientcontext/ClientContentExtractionFilter.java
```diff
@@ -23,6 +26,14 @@
 
     @Override
     protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain) throws ServletException, IOException {
+        log.info(String.format("Request before ClientContentExtractionFilter: %s",request.toString()));
+        Enumeration gwwHeaderNames = request.getHeaderNames();
+        log.info("Logging headers...");
+        while (gwwHeaderNames.hasMoreElements()) {
+            String key = (String) gwwHeaderNames.nextElement();
+            log.info("Header {}: {}", key, request.getHeader(key));
```
jrmihalick, line 34: Late to the party on this PR, but same question as on other one.  Should probably remove or downgrade this logging to debug.

PR #1723, opened 2021-11-03
Feature/customer accounts

stojankz:
### Summary
Openpos-framework PR for customer account and customer subscription logic being pulled from 1.12 into master. List of PRs cherrypicked and pulled in can be found here: 

https://docs.google.com/spreadsheets/d/1fd6ig9cofc6fXI2nlEa2RGqB61MAx2Q0RAbQVhNsF3Y/edit?usp=sharing

### Screenshots
Commerce Demo:
customer-subscription-demo

Commerce Demo Fashion:
customer-subscription-demo-fashion


chenson42 (dismissed):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/enrollment-confirmation-dialog/enrollment-confirmation-dialog.html
```diff
@@ -0,0 +1,37 @@
+<div class="dialog-outer">
+    <app-dialog-header></app-dialog-header>
+    <div class="dialog-body" responsive-class>
+        <app-content-card>
+            <div *ngIf="screen.message && screen.programCopy">
+                <div [innerHTML]="safeProgramCopy"></div>
+                <div class="message">
+                    <app-icon *ngIf="screen.checkIcon" [iconName]="screen.checkIcon"
+                              [iconClass]="'material-icons mat-64'"></app-icon>
+                    <div [innerHTML]="screen.message"></div>
```
kdelorey, line 10: Thoughts on the `screen.message` (and similar) being markdown instead of RAW html and rendered with the pipe? This will likely result in an easier time with any style and for any really complex things that markdown can't handle you can still embed HTML into the markdown.

kdelorey (dismissed):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/enrollment-confirmation-dialog/enrollment-confirmation-dialog.html
stojankz: does the markdown-formatter pipe work if screen.message is html? I'm not entirely sure behind the reasoning but I believe with some of the details shown on a lot of these dialogs, the idea was to have them be completely customizable in how they look and so that message is coming from a database table like cust_customer_program_copy

r-cochran (dismissed):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/enrollment-confirmation-dialog/enrollment-confirmation-dialog.html
r-cochran: that is correct, there was a desire from the client for the ability to fully create the content in the various copy locations (program / plan) for content/styling/layout. 

If the markdown pipe will let all the content through as it works today and do replacements for anything that is markdown then I don't see an issue with making this change. Though I feel like if the person making the copy later on accidentally hits on any of the markdown syntax they will have a headache debugging it.

kdelorey: It _should_ allow you to a mix of markdown and html, but if I'm behind honest I don't have the details of that on me. I assumed customizing the HTML with a bunch of styles would be a pain point that would be alleviated with markdown. Its probably not worth throwing wrenches into the solution at the moment however, maybe something we explore down the road. Lets leave it alone for now.

jchirico8 (changes requested):

jchirico8 (approved):

r-cochran (approved):

jchirico8 (approved):

PR #1729, opened 2021-11-04
Only invoke endpoint filter on objects of the sought type

jrmihalick:
### Issues Fixed
PDPOS-6838
_Error messages appearing in CO commerce log_ 

### Summary
- Previously we were visiting all collections of an outgoing endpoint response rather than only visiting objects of the matching target class defined by the endpoint filter.  This was causing the error below.  Change the logic in the `ObjectFinder` to only visit objects whose type matches the target type.
- With this change, the best we can do for setting the `parent` value in the visitor is to set it to a reference to the parent collection for the case when objects of the target type are in a collection.
- Fixes this exception:
```
org.jumpmind.pos.service.PosServerException: Failed to invoke filter method public org.jumpmind.pos.customer.model.CustomerModel org.jumpmind.pos.petco.customer.CustomerEndpointFilter.groupIdPrivateLabelToPcc(org.jumpmind.pos.customer.model.CustomerModel)
at org.jumpmind.pos.service.filter.EndpointFilterManager.invokeFilter(EndpointFilterManager.java:234)
at org.jumpmind.pos.service.filter.EndpointFilterManager.lambda$filterResponse$0(EndpointFilterManager.java:192)
at org.jumpmind.pos.util.ObjectFinder.searchRecursive(ObjectFinder.java:120)
at org.jumpmind.pos.service.filter.EndpointFilterManager.filterResponse(EndpointFilterManager.java:190)
at org.jumpmind.pos.service.filter.EndpointFilterManager.filterResponse(EndpointFilterManager.java:170)
at org.jumpmind.pos.service.EndpointInvoker.invokeStrategy(EndpointInvoker.java:367)
at org.jumpmind.pos.service.EndpointInvoker.invoke(EndpointInvoker.java:354)
```


openpos-service/src/main/java/org/jumpmind/pos/service/filter/EndpointFilterManager.java
```diff
@@ -188,16 +188,14 @@ protected Object filterResponse(EndpointInvocationContext context, EndpointFilte
         } else {
             ObjectFinder<?> finder = new ObjectFinder<>(filterTemplate.getOutputType());
             finder.searchRecursive(argument, (parentObject, targetObject, field) -> {
```
jrmihalick, line 190: Now that we are only visiting objects of the target type, I don't see a need to check `parentObject == argument`.

mmichalek (approved):

PR #1740, opened 2021-11-10
Add scrolling to sale total panel for some resolutions

mdruzhin:
### Issues Fixed

[PCOM-5440](https://jira.ae.com/browse/PCOM-5440)

### Summary

When we have some specific resolution (for example 1366x768) checkout button is not shown fully and scrolling is not available for this one. I added an opportunity to scroll. I checked other resolutions and it looks good with these changes.

### Demonstration

recording (6)

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/sale/sale.component.scss
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/sale/sale.component.scss
@@ -78,6 +78,7 @@
             }
 
             app-sale-total-panel {
+                overflow-y: auto;
                 grid-area: 1/3/1/3
             }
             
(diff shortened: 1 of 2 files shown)


kdelorey:
I kind of wonder if we should instead `display: flex` some of these things or consider different font sizes before resorting to scrolling? @blubaugha wacha think?

mdruzhin:
For this situation just resolution too low for horizontal orientation so we just don't have more place to showing it without scrolling.

kdelorey:
I guess I was referring to some of the items like Tax, Discounts and Subtotal; it feels like we could fit them into two columns or something instead and if need by we could change the font sizes so they aren't so bold and use up a ton of valuable space. This goes for Amount Due too, the font sized could be lessened some on mobile layouts in order to save some space.

mdruzhin:
It could help to resolve this issue without scrolling. @chenson42 What do you think about this solution?

mdruzhin:
@kdelorey @chenson42  I took a look on styles for this resolution and it's not considered as mobile device. This issue is reproduced for any resolutions which have height is less than 830px (with linked loyalty customer and horizontal orientation). We can add styles (as you described) for these resolutions as well.

chenson42:
I don't understand why this is an issue at all.  Seems like the user who entered the issue just has a super low resolution set. 
 None of the form factors that we deploy to have this issue.   Should we be doing anything about this right now?

mdruzhin:
@chenson42 I agree with you, I didn't have this issue with resolutions which we are using. I'm not sure if this change is needed or not.

mdruzhin:
This issue have been discussed and will be created other PR for this one.

PR #1743, opened 2021-11-12
Adjustments to get Petco onto master

brandon-lewis:
### Summary
Minor updates to support commerce changes in order to get Petco onto master.


openpos-devices/src/main/java/org/jumpmind/pos/devices/service/model/GetDeviceRequest.java
```diff
@@ -11,4 +11,5 @@
 @NoArgsConstructor
 public class GetDeviceRequest {
     private String deviceId;
+    private String appId;
```
stojankz, line 14: I believe appId was purposefully removed in master. I don't think we should be adding it back

stojankz (changes requested):

openpos-devices/src/main/java/org/jumpmind/pos/devices/service/model/GetDeviceRequest.java
jrmihalick: It was removed from the dev_device_auth table, so perhaps the removal from GetDeviceRequest is related.  I don't think removing it will cause a backwards compatibility issue since the GetDeviceRequest is not passed remotely from one commerce instance to another.

jrmihalick (approved):
Look ok pending adjustments for removal of the appId on GetDeviceRequest.

jrmihalick (approved):

stojankz (approved):

PR #1746, opened 2021-11-15
Add support for javapos printers

chenson42:
### Summary
Add support for javapos printers


openpos-peripheral/src/main/java/org/jumpmind/pos/print/EscpPOSPrinter.java
```diff
@@ -465,6 +386,7 @@ public String readMicr() {
         }
     }
 
+    // TODO should this be beingInsertion
```
jchirico8, line 389: I don't think this should be _being_ anything
```suggestion
    // TODO should this be beginInsertion
```

jchirico8 (approved):

PR #1747, opened 2021-11-15
WIP - Changes for SFS/BOPIS order components

eddielegg-le:
### Summary
Openpos changes to support updates to order UI. 

### Database Changes
List database changes that apply and note their ramification.  If this section applies then add the label _databasechange_ to your pull request.

### Parameter Changes
Have parameters been added, renamed or removed?  If this section applies then add the label _parameterchange_ to your pull request.

### Screenshots
Screenshots go here.

### DIFF
--- a/openpos-flow/src/main/java/org/jumpmind/pos/core/ui/data/OrderSummary.java
+++ b/openpos-flow/src/main/java/org/jumpmind/pos/core/ui/data/OrderSummary.java
@@ -27,6 +27,7 @@ public class OrderSummary implements Serializable {
     private String amountDue;
     private String itemCountIcon;
     private TimeUnitLabels timeUnitLabels;
+    private String orderTypeCode;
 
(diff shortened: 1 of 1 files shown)


openpos-flow/src/main/java/org/jumpmind/pos/core/ui/data/OrderSummary.java
```diff
@@ -216,6 +218,14 @@ public TimeUnitLabels getTimeUnitLabels() {
     public void setTimeUnitLabels(TimeUnitLabels timeUnitLabels) {
         this.timeUnitLabels = timeUnitLabels;
     }
+
```
stojankz, line 221: Could we actually get rid of all the getters and setters and update the OrderSummary to use lombok?

stojankz (changes requested):

PR #1756, opened 2021-11-17
Don't Squash the Form!

blubaugha:
### Issues Fixed
[PCOM-5509](https://jira.ae.com/browse/PCOM-5509)

### Summary
This fixes a style problem with the form in mobile devices, on the customer display, where the form content would be squashed when the mobile keyboard becomes visible.

The fix moves the scrollable area so that you have to scroll to the bottom of the page to see the form buttons. Previously, the form buttons were always visible, but that ended up causing this bug.

### Screenshots
customer-display-form


kdelorey:
I'll try this on the iPad, I don't suspect any problems... one minute.

kdelorey:
image

Seems better to me, not sure how much more improvement we could make.

kdelorey (approved):

PR #1758, opened 2021-11-18
Expansion panel component

eddielegg-le:
### Summary
Custom expansion panel component with content projection for header and content containers:

https://user-images.githubusercontent.com/74971030/142483687-942a042e-3d17-4349-bd43-a882f6eca722.mov


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/expansion-panel/_expansion-panel-theme.scss
```diff
@@ -0,0 +1,19 @@
+@mixin expansion-panel-theme($theme){
+
+  $app-primary: mat-color(map-get($theme, primary));
+  $app-accent: mat-color(map-get($theme, orders));
```
stojankz, line 4: for the generic component this accent color should probably be 
```suggestion
$app-accent: mat-color(map-get($theme, accent));
```

eddielegg-le: Yep good call

stojankz (approved):

PR #1773, opened 2021-11-29
Sale Screen layout Fix

r-cochran:
### Summary
fixed a misplaced div that caused the sale screen to not display all of the cart details and checkout functionality

### Screenshots
before
image

after
image


stojankz:
hm.. did you test this change running commerce demo at all? I don't see this happen running demo fashion so I'm a little worried to update the openpos component without thorough testing
Screen Shot 2021-11-29 at 1 33 20 PM

chenson42 (approved):

r-cochran:
> 
I had not, but good call out. It looks to display correctly with the change in this PR.

image

If I swap back to master this is what I get

image

I think this might have been a mistake introduced with [this change](https://github.com/JumpMind/openpos-framework/commit/2f768f2f85a43ea4025971f853f5bce4a2e6e48d)

stojankz:
> > 
> 
> I had not, but good call out. It looks to display correctly with the change in this PR.
> 
> image
> 
> If I swap back to master this is what I get
> 
> image
> 
> I think this might have been a mistake introduced with [this change](https://github.com/JumpMind/openpos-framework/commit/2f768f2f85a43ea4025971f853f5bce4a2e6e48d)

ah okay, looks like it broke today and I don't have that change pulled in which is why I did not see it.

stojankz (approved):

chenson42:
Yeah.  That is why I so quickly approved it.

PR #1784, opened 2021-12-02
Cache i18n requests at the client to reduce network noise

chenson42:
### Summary
Cache i18n requests at the client to reduce network noise


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/locale.service.ts
```diff
@@ -43,19 +44,26 @@ export class LocaleService {
     }
 
     getString(base: string, key: string, args?: any[]): Observable<string> {
-        const url = `http${this.personalization.getSslEnabled$().getValue() ? 's' : ''}` +
+        const cacheKey = base + ':' + key + (args ? args.join(':') : '');
+        const text = this.textCache.get(cacheKey);
+        if (text) {
+            console.log(`using cached i18n value for ${cacheKey} = ${text}`);
```
jrmihalick, line 50: This should probably be console.debug?  Which doesn't show up by default in the console and might get filtered out when we are sending log messages from client to server?

jrmihalick (approved):

kdelorey (approved):

PR #1788, opened 2021-12-03
Add auto personalization using URL from enterprise config

jrmihalick:
### Issues Fixed
PDPOS-5212
mPOS (IAC/Vetco/Mobile Grooming) Automate personalization for the TC-52 / TC-57

### Summary
- Add the concept of an `IEnterpriseConfig`, `EnterpriseConfigStartupTask`, and `EnterpriseConfigService` for retrieving application configuration information from a configuration that is supplied by the deployment platform.  For example, [Airwatch Custom Configuration](https://docs.vmware.com/en/VMware-Workspace-ONE-UEM/index.html), an [Android Managed Configuration](https://developer.android.com/work/managed-configurations), or an [iOS Managed Configuration](https://support.apple.com/guide/mdm/welcome/web).
- The `IEnterpriseConfig` interface provides a mechanism like the existing `IPlatformPlugin` to make it possible to provide one or more implementations for accessing enterprise configuration data.  This pull request includes an `AirwatchCordovaPlugin` class which uses the `JMCAirwatchPlugin` (a new JumpMind cordova plugin, which will be in a separate PR), if present, to access Airwatch Custom Configuration data through the Airwatch SDK.  
- Access to the `IEnterpriseConfig` configuration data is provided through the `EnterpriseConfigService`. If there are multiple implementations of `IEnterpriseConfig` objects present at runtime, the `EnterpriseConfigService` merges all individual configurations present into a single configuration.  `IEnterpriseConfig` objects are loaded in the order they are specified in the `core.module.ts` and in a "last one wins" manner.  So, if two configurations contain the same parameter, the parameter loaded from the last configuration wins.  
- The `EnterpriseConfigStartupTask` runs very early in the set of startup tasks to simply run the initialization of the `EnterpriseConfigService`, which loads all of the configuration data from the individual `IEnterpriseConfig` objects.  If no `IEnterpriseConfig` objects are available at runtime, the resultant configuration provided by the `EnterpriseConfigService` simply is empty.
- I've modified the Auto-personalization algorithm in `AutoPersonalizationStartupTask` to first attempt personalization with the configuration provided by the `EnterpriseConfigurationService`.  Specifically, it looks for a parameter in the configuration named `jmc_autoPersonalizeUrl`.  If no parameter by that name is found in the configuration, the algorithm falls back to attempting personalization via Zero conf, as it did previously.  If a parameter named `jmc_autoPersonalizeUrl` is found in the configuration, a call is then made to the Auto personalization endpoint and the entire enterprise configuration data is supplied in the request.  This allows for some flexibility in the server side endpoint implementation to do custom work with the configuration data supplied.  Petco, for example, does not use a device name in their Android devices, but they do have a device serial number, which is available in their enterprise configuration. The device serial number can then be used in a petco-specific override of the auto personalization endpoint to look up the device personalization info.
- I've changed the existing openpos `GetDevicePersonalizationModelEndpoint`, to accept a request object instead of just a deviceName.  The new request object now has both `deviceName` and `additionalAttributes` to support passing over arbitrary configuration info on the auto personalization request.


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/cordova-plugins/airwatch-cordova-plugin.ts
```diff
@@ -0,0 +1,93 @@
...
+                            } else if (eventName === 'initFailure') {
+                                observer.error(`Airwatch SDK failed to initialize. Reason: ${eventData ? (eventData.error ? eventData.error : JSON.stringify(eventData)) : '?'}`);
+                            }
+                        });
+                    } else {
+                        observer.complete();
+                    }
+                })
+            ).subscribe();
+        });
```
kdelorey, line 41: Wrapping this `.subscribe()` within a `new Observable<string>(...)` is really confusing to me. Particularly the `.map()` portion. Instead doing the intention of `map()` which is to coerce the stream value, it's actually operating on the `observer`. `tap()` or preferably just filling out the `.subscribe({ next: (event) => {} })` would be be better suited for how you're currently using it.

Something kinda like this makes more sense to me simply because I can grasp what its doing from the `.pipe()` operations.

```ts
initialize(): Observable<string> {
   if (this.initialized) {
       return of();
   }

   return this.cordova.Service.onDeviceReady.pipe(
       filter(m => m === 'deviceready'),
       tap(() => this.jmcAirwatchPlugin = window. JMCAirwatchPlugin),
       switchMap(() => new Observable(observer => {
           this.jmcAirwatchPlugin.setSDKEventCallback((eventName, eventData) => {
              if (eventName === 'initSuccess') {
                  this.initialized = true;
                  observable.next('Airwatch SDK successfully initialized');
                  observable.complete();
              } else if (eventName == 'initFailure') {
                  observer.error('the error message');
              }
           });
           
           // todo: this is just what I assume you do to unsubscribe, might need alterations
           return () => this.jmcAirwatchPlugin.setSDKEventCallback(undefined);
       })
   )
}
```

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/cordova-plugins/airwatch-cordova-plugin.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    private isInitialized(): boolean {
+        if (! this.initialized) {
+            console.error('AirwatchCordovaPlugin is not yet initialized, cannot complete requested operation');
+            return false;
+        }
+        return true;
+    }
+
+    getConfiguration(): Observable<any> {
```
kdelorey, line 64: Do we not have an idea as to what the configuration looks like? Can we avoid `any`?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/cordova-plugins/airwatch-cordova-plugin.ts
```diff
@@ -0,0 +1,93 @@
...
+        return new Observable<any>( observer => {
+            if (! this.isInitialized()) {
+                observer.complete();
```
kdelorey, line 67: It seems like returning an `.error()` downstream is more appropriate?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/auto-personalization-startup-task.ts
```diff
@@ -109,28 +137,47 @@ export class AutoPersonalizationStartupTask implements IStartupTask {
...
+                    return EMPTY;
+                }
+            }),
+            // This is causing a ts-lint error, but I cannot figure out why
+            catchError(e => {
+                console.log('[AutoPersonalizationTask] error during attempt to Auto personalize');
+                observer.next(`Error during auto-personalization to host '${url}': ${JSON.stringify(e)}`);
+                this.logPersonalizationError(e);
+                throw(e);
+            })
```
kdelorey, line 184: `catchError` has a very specific purpose to catch the error replacing it with a new sequence. You should not `throw` an exception from here. `tap()`'s error argument would be better suited here or optionally instead of `throw(e)` you can `return throwError(e);` and `import { throwError } from 'rxjs';`

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/auto-personalization-startup-task.ts
```diff
@@ -109,28 +137,47 @@ export class AutoPersonalizationStartupTask implements IStartupTask {
...
+                        info.sslEnabled).pipe(
+                            map(msg => {
+                                observer.next(msg);
+                                observer.complete();
+                                stopSignal$.next();
+                            }),
+                            takeUntil(stopSignal$),
+                            catchError(e => {
+                                this.logPersonalizationError(e);
+                                throw e;
```
kdelorey, line 171: same as `@178` comment

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/enterprise-config/enterprise-config.service.ts
```diff
@@ -0,0 +1,126 @@
...
+                const inits = this.enterpriseConfigs.map( cfg => {
+                    return concat(
+                        of(`Initializing ${cfg.name()}`),
+                        cfg.initialize()
+                    );
+                });
+
+                merge( ...inits ).subscribe( {
+                    next: m => messages.next(m),
+                    error: e => messages.error(e),
```
kdelorey, line 48: if there are multiple `inits` and has an error, it will cause this entire stream to fail with an `error`. Did you want that, or do you want to omit it from use somehow?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/enterprise-config/enterprise-config.interface.ts
```diff
@@ -0,0 +1,32 @@
+import {Observable} from "rxjs";
+
+export interface IEnterpriseConfig {
+
+    /**
+     * Unique name to identify this configuration
+     */
+    name(): string;
```
kdelorey, line 8: As a method I would expect this to be a verb, like `getName(): string`. In this case I would expect it to be a readonly field

```suggestion
    readonly name: string;
```

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/cordova-plugins/airwatch-cordova-plugin.ts
jrmihalick: I appreciate the feedback here.  I struggle to get these observables right, it drives me pretty crazy.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/cordova-plugins/airwatch-cordova-plugin.ts
jrmihalick: I'll review that

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/cordova-plugins/airwatch-cordova-plugin.ts
jrmihalick: I get your point, but I don't know what to do here since the configuration could vary per retailer.  Currently I am just passing the configuration through to the server side which has more flexibility for retailer-specific handling of the configuration info.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/auto-personalization-startup-task.ts
jrmihalick: Thanks, I will see if I can improve this.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/enterprise-config/enterprise-config.interface.ts
jrmihalick: Sure, I'll adjust this.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/enterprise-config/enterprise-config.service.ts
jrmihalick: I do really want to omit the one with an error, yeah.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/cordova-plugins/airwatch-cordova-plugin.ts
kdelorey: I know they aren't used a ton in commerce, so I don't know what your experience with them are, but the advice I gave @chenson42 the other day was to think of the `pipe()` operators like the transforms on Java `Stream`. They're often the very same idea and often share the same names too.

jrmihalick: I understand them conceptually, but the struggle is in knowing the api and knowing which operators to use and when.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/cordova-plugins/airwatch-cordova-plugin.ts
kdelorey: If its unavoidable then there's nothing wrong here. I just wanted to make sure.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/enterprise-config/enterprise-config.interface.ts
jrmihalick: I'm not going to make this change because `platform-plugin.interface.ts` already has a `name()` method and I purposefully want to have the flexibility for the implementations of the interface to both a IPlatformPlugin and a IEnterpriseConfig.  typescript will not allow me to have both a `name` attribute and a `name()` method.

kdelorey (approved):

PR #1804, opened 2021-12-10
Plugin dependecies missing and wrong usage of plugin name

shiljo-litmus7:
Issues Fixed
Dependencies were not provided for the plugin being injected. Also the name which Cordova Platform binds to window object was not correct as per the plugin definition xml

Summary
Fixing the Cordova plugin injection issue


jrmihalick (approved):
Approve, as long as you end up keeping this plugin.  If you do keep it the cordova plugin implementation you are using, please add a comment to the class that mentions the name of the actual cordova plugin that is required.

PR #1809, opened 2021-12-14
Using a different Cordova plugin for exiting the app

shiljo-litmus7:
Issues Fixed
The Cordova plugin which we initially used to Exit the App was not cleanly closing the app. Found a different plugin that force closes the App. 

Summary
Since the new Cordova plugin will bind to a different window property, we need to accommodate that change in openpos plugin


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/cordova-plugins/auto-exit.ts
```diff
@@ -40,7 +40,8 @@ export class ExitAppPlugin implements IPlatformPlugin {
                 filter(m => m === 'deviceready'),
                 take(1),
                 tap(() => {
-                    this.ExitAppCordovaPlugin = window['cordova'].plugins['exit'];
+                    let objHold:any = window.navigator;
+                    this.ExitAppCordovaPlugin = window.navigator?objHold.app:false;
```
jrmihalick, line 44: Could you add the name of the plugin required for this to work?  `cordova-plugin-app-exit`?

jrmihalick (approved):
See my comment

PR #1812, opened 2021-12-15
Shopping Pass Tender Screen Next button disabled

chadhav-ae:
### Issues Fixed
Fix to enable next button in shopping pass when manual enter the expiry date.


chenson42:
@stojankz @jchirico8 @blubaugha Can you please review?

kdelorey:
Let me look at your issue @chadhav-ae, the changes here feel a bit too hacky.

chadhav-ae:
> Let me look at your issue @chadhav-ae, the changes here feel a bit too hacky.

Thanks Kevin, i appreciate it. Yes i believe that too it is not the right way. This is working but it is not the correct way.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/prompt-form-part-card/prompt-form-part-card.component.ts
```diff
@@ -140,4 +149,19 @@ export class PromptFormPartCardComponent extends ScreenPartComponent<PromptFormP
        }
     }
 
+    onPromptDateChange(event): void {
+        const dateregex = '[0-9]{2}[/][0-9]{2}[/][0-9]{3}[_]$';
+        const dateregex2 = '[0-9]{2}[/][0-9]{2}[/][0-9]{4}$';
```
kdelorey, line 154: You can do this with a single regular expression. But also you anchor the end but not the beginning, so you would allow any input before this tring and it would still "match". I would expect something like `^\d{2}\/\d{2}\/(?:\d{3}_)|\d{4}$`

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/prompt-form-part-card/prompt-form-part-card.component.ts
```diff
@@ -124,8 +126,15 @@ export class PromptFormPartCardComponent extends ScreenPartComponent<PromptFormP
         if (this.promptFormGroup.valid && this.options.nativeElement.children.length === 0) {
             const payload = this.promptFormGroup.value[this.inputControlName];
             const payload2 = this.promptFormGroup.value[this.inputDateControlName];
-            if (this.screenData.actionButton) {
-                this.doAction({ action: this.screenData.actionButton.action }, {creditCardNumber: payload, expiryDate: payload2});
+            const dateregex = '[0-9]{2}[/][0-9]{2}[/][0-9]{4}$';
+            if (!payload2.match(dateregex) && this.screenData.isExpiryDateEnabled) {
```
jchirico8, line 130: Is there not a regex on the date field already? In order to get here, you've confirmed that the whole form group is valid. It would be better if the regex was a pattern validator on the field itself so angular would have it marked as invalid

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/prompt-form-part-card/prompt-form-part-card.component.ts
```diff
@@ -140,4 +149,19 @@ export class PromptFormPartCardComponent extends ScreenPartComponent<PromptFormP
...
 
+    onPromptDateChange(event): void {
+        const dateregex = '[0-9]{2}[/][0-9]{2}[/][0-9]{3}[_]$';
+        const dateregex2 = '[0-9]{2}[/][0-9]{2}[/][0-9]{4}$';
+        const val = event.target.value;
+        if (val.match(dateregex) || val.match(dateregex2)) {
+            this.promptFormGroup.get('promptDateInputHiddenDateControl').setValidators(null);
+        } else {
+            this.promptFormGroup.get('promptDateInputHiddenDateControl').setValidators([Validators.required]);
+        }
```
jchirico8, line 160: This seems weird to me. Why would the value in the field change how we validate it? This looks like after the first character is entered, we remove the minDate validator you added before because it overwrites validators with only required

kdelorey:
I think we should close this in favor of #1823. Reopen if you feel otherwise.

PR #1815, opened 2021-12-20
Startup 2.0

kdelorey:
# Summary
This proposes new changes to the startup sequence in order to make it easier to troubleshoot, unit test, and comprehend. Previously the startup system worked by chaining together really noise observable sequences and their completion signaled their success. This worked great but relied on heavy knowledge of RxJS operators to properly understand and even with advanced knowledge, some tasks were very difficult to comprehend and subsequently troubleshoot.

Overall it was a really chatty stream as the stream contained splash screen messages, most often they were too verbose containing log messages. This chatty-ness made it difficult to do proper unit testing on startup tasks.

## Interface
The new `StartupTask` interface contains only a single `execute()` method and is permitted to return a synchronous `void`, an asynchronous `Promise<void>` or an `Observable<any>`. The task runner will understand each type and act accordingly. It will be suggested that new tasks stick to `void` or `Promise<void>` when possible.

Implementations of `StartupTask` should always be a `@Injectable()` in order to properly support Dependency Injection. If it is not, task execution will fail.

The name of the task is automatically inferred from the name of the class.

## Startup Sequence
The priory scheme of tasks as been dropped in favor of just globally listing them in sequence order. In order to add a new task, you should add it to the `startupSequence` in `startup-sequence.ts`. This list accepts injection tokens or combinators.

### `all()` Combinator
The `all(...)` combinator requires all of the listed task to complete successfully or execution of the combinator fails. The combinator fails at the first error.

### `oneOf()` Combinator
The `oneOf(...)` combinator will run tasks sequentially and completes when the first task finishes successfully. If all tasks fail execution, the combinator fails to execute.

### `optional()` Combinator
The `optional()` combinator will run the contained task, but succeeds on failure and success.

### `withSplashMessage()` Combinator
The  `withSplashMessage()` will simply wrap a task and during execution it will set the displayed splash screen message to a user friendly string. When execution is finished, the splash screen message is popped off.


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/startup-controller.service.ts
```diff
@@ -0,0 +1,43 @@
+import { Injectable, Injector } from '@angular/core';
+import { startupSequence } from './startup-sequence';
+
+export type StartupControllerState = true | 'in-progress' | false | 'error';
```
jchirico8, line 4: Would this work the same?
```suggestion
export type StartupControllerState = boolean | 'in-progress' | 'error';
```

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/startup-controller.service.ts
```diff
@@ -0,0 +1,43 @@
...
+    };
+
+    constructor(private _injector: Injector) { }
+
+    async beginStartupSequence(): Promise<StartupControllerStatus> {
+        if (this._status.completed !== false) {
+            throw new Error('startup sequence already initiated');
+        }
+
+        this._status = { completed: 'in-progress' };
```
jchirico8, line 30: This might just be preference but I'm thinking why initialize the property above if you'll overwrite it with a new object instead of just changing the status
```suggestion
        this._status.completed = 'in-progress';
```

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/startup-controller.service.ts
```diff
@@ -0,0 +1,43 @@
...
+
+        try {
+            await startupSequence.executeTask(this._injector);
+        } catch (e) {
+            this._status = { completed: 'error', error: e };
+            return;
+        }
+
+        this._status = { completed: true };
```
jchirico8, line 39: Same thing here. Why create a whole new object each time
```suggestion
        this._status.completed = true;
```

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/startup-task.ts
```diff
@@ -0,0 +1,215 @@
...
+class OptionalStartupTaskCombinator extends StartupTaskCombinator {
+  constructor(private _task: StartupTaskLike) {
+    super();
+  }
+
+  async executeTask(injector: Injector): Promise<void> {
+    const result = await super.performTask(injector, this._task);
+
+    if (!result) {
+      console.log(`startup task '${getTaskName(this._task)}' failed to execute; ignoring optional task`);
```
jchirico8, line 114: does this log need the task name in it? The onfOf failure message doesn't I assume because there is a log in performTask that would tell you what task is running right before the failure

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/startup-controller.service.ts
kdelorey: Yeah it should; not sure why I did it this way.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/startup-task.ts
kdelorey: I think this was remnants of me debugging. I think I do want to leave it will always successfully execute whereas `oneOf()` can fail. So it might be nice to know that underneath that the task actually failed, so I think I'd like to change this to a `debug` level message. I do think `oneOf` logging could be improved too. I'll look into it.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/startup-controller.service.ts
kdelorey: I actually prefer the state as one whole immutable object (which I failed to actually do). Changes to to the object should be detected as a whole not as two seperate changes. It's also a little bit for safety. Changing the state should force you to intentionally attach an error or not.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/tasks/zero-conf-personalization.startup-task.ts
```diff
@@ -0,0 +1,103 @@
...
+import { CONFIGURATION } from '../../../configuration/configuration';
+import { AutoPersonalizationParametersResponse } from '../../personalization/device-personalization.interface';
+import { PersonalizationService } from '../../personalization/personalization.service';
+import { Zeroconf, ZeroconfResult } from '../../zeroconf/zeroconf';
+import { StartupTask } from '../startup-task';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class ZeroConfPersonalizationStartupTask implements StartupTask {
```
jrmihalick, line 12: We're going to need an Auto Personalization base class that does that part of calling the server to get the personalization parameters.  That could then be extended by this task and another that I will need to port over from 1.13 which uses the EnterpriseConfig to personalize.  The only difference really between those two tasks is where the URL used for attempting to get the auto personalization params is coming from.  While you are in here doing this, might as well prepare for it now and set up the base class now.  😄

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/startup-sequence.ts
```diff
@@ -0,0 +1,36 @@
...
+import { CapacitorHideStatusbarStartupTask } from './tasks/capacitor-hide-statusbar.startup-task';
+import { InitializePersonalizationStartupTask } from './tasks/initialize-personalization.startup-task';
+import { ManualPersonalizeStartupTask } from './tasks/manual-personalization.startup-task';
+import { PlatformDiscoveryStartupTask } from './tasks/platform-discovery.startup-task';
+import { PlatformPluginsStartupTask } from './tasks/platform-plugins.startup-task';
+import { SavedSessionPersonalizationStartupTask } from './tasks/saved-session-personalization.startup-task';
+import { SessionConnectStartupTask } from './tasks/session-connect.startup-task';
+import { ZeroConfPersonalizationStartupTask } from './tasks/zero-conf-personalization.startup-task';
+
+export let startupSequence = all(
```
jrmihalick, line 12: This is a big improvement, better than having an `order` attribute.  👍

jrmihalick: Plus we'll be able to add new combinators.  Nice.

jrmihalick (changes requested):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/tasks/audio.startup-task.ts
```diff
@@ -0,0 +1,31 @@
...
+})
+export class AudioStartupTask implements StartupTask {
+    constructor(
+        private audioRepositoryService: AudioRepositoryService,
+        private audioService: AudioService,
+        private audioInteractionService: AudioInteractionService
+    ) { }
+
+    async execute(): Promise<void> {
+        console.debug('begin preloading audio configuration');
```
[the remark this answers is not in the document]
kdelorey: There are some side effects to consider. Automatically augmenting the console logging will cause the little portion of the chrome devtools that shows where the it was logged from to be overwritten with something useless. This is probably acceptable since you want to include the task name in the log message. It will also impact everything logging from services and such while inside of a task. It would look something like:

```
[SomeStartupTask] About to call some service
[SomeStartupTask] Okay I'm in the service now doing stuff.
```

Not sure if that's the desired effect you want?

Another option to avoid all these side effects is to manually place this information within log message. This would be pretty tedious and annoying to do in my opinion, especially considering the information is already present between begin and end log blocks.

We could also roll a logger that is understands some scope?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/tasks/zero-conf-personalization.startup-task.ts
kdelorey: Sure, I also realize that I think I accidently forgot the personalize with query params so I'll do that too.

kdelorey: Done, but you'll have to modify it to support your enterprise stuff which doesn't appear to be in master.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/tasks/audio.startup-task.ts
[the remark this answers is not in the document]
kdelorey: image

I did the first option for now, we can improve it later if you want.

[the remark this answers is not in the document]
jrmihalick: @kdelorey Looks good, thanks.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/tasks/zero-conf-personalization.startup-task.ts
jrmihalick: @kdelorey Yep, no problem

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/startup/tasks/auto-personalization.startup-task.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    abstract execute(): void | Promise<void> | Observable<any>;
+
+    protected async personalize(deviceName: string): Promise<void> {
+        let info: AutoPersonalizationParametersResponse;
+
+        try {
+            info = await this.personalization.getAutoPersonalizationParameters(
+                deviceName,
+                CONFIGURATION.autoPersonalizationServicePath
```
jrmihalick, line 19: Why hard-code read this out of configuration instead of passing in as a parameter?  When I port over the EnterpriseConfig changes, am I supposed to directly set `CONFIGURATION.autoPersonalizationServicePath` with the value I read from the Enterprise config?

kdelorey: This is just what existed before and I didn't really give it much thought. I think @eddielegg-le or somebody added it to support some other way of auto personalizing. Wasn't really sure what your needs are so I just ripped a possibly common chunk out into a base class as requested. You might just modify it to fit your needs.

Doesn't your enterprise variant use the same `personalizeMe` endpoint? Is that not what this configuration variable points to?

jrmihalick: Yes, it does use the same personalizeMe endpoint, but the actual URL of the endpoint will vary depending on the environment.  I use the enterprise config to read that URL value.  If you prefer, I can refactor later, but I would probably just change this task to pass in the URL as a param.

eddielegg-le: Yeah that was what we decided on for Petco for the zero conf personalization. Not even sure it's used now?

kdelorey: I would just refactor it to your needs later. I'm also partially biased because my hands are tied up with AEO stuff currently so I probably won't touch this PR again for a long while.

jrmihalick (approved):

PR #1817, opened 2021-12-20
Feature/sfs packing

stojankz:
### Summary
openpos updates made for order packing. Changes include removing box-shadow, fix to icons scss bug where icons were not being resized correctly to be smaller when in mobile view, stopping propagation on the icon-button click, and adding necessary icon svgs.


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/icon-button/icon-button.component.ts
```diff
@@ -14,7 +14,8 @@ export class IconButtonComponent {
 
     @Output() buttonClick = new EventEmitter();
 
-    clickFn(): void {
+    clickFn(event: any): void {
         this.buttonClick.emit();
```
eddielegg-le, line 18: Not a big deal right now, but if you're passing the event into the function, might as well pass it along in the emitter too.

stojankz: sure thing

eddielegg-le (approved):

jchirico8 (approved):

PR #1820, opened 2021-12-22
Enhance personalization and the ITagProvider to allow access to all device attributes

jrmihalick:
### Summary
- Adds `IDeviceAttributes` interface in order to keep `ITagProvider` in `openpos-persist` project.  `DeviceModel` implements the `IDeviceAttributes` interface so that the device properties can be provided to `ITagProvider` implementations.
- This is in preparation for dynamically setting the `device_type` tag during Petco personalization
- Clean up a couple of lint warnings


openpos-devices/src/main/java/org/jumpmind/pos/devices/model/DeviceModel.java
```diff
@@ -59,6 +58,8 @@
     @Builder.Default
     private String deviceMode = DEVICE_MODE_DEFAULT;
 
+    private List<DeviceParamModel> deviceParamModels;
```
jrmihalick, line 61: I just moved this from the bottom of the class def

jrmihalick:
@chenson42, @mmichalek  : Could I get a review on this since it will eventually be coming to master?

openpos-persist/src/main/java/org/jumpmind/pos/persist/ITagProvider.java
```diff
@@ -1,10 +1,10 @@
 package org.jumpmind.pos.persist;
 
-import org.springframework.core.env.Environment;
+import org.jumpmind.pos.util.model.IDeviceAttributes;
 
 
 public interface ITagProvider {
 
-    public String getTagValue(String deviceId, String businessUnitId, String tagName);
+    String getTagValue(IDeviceAttributes deviceModel, String tagName);
```
chenson42, line 8: maybe change deviceModel to attributes

jrmihalick: Whoops, yeah

chenson42 (approved):

PR #1823, opened 2021-12-23
Experimental Prompt w/ Multiple Inputs and Working Form Validation

kdelorey:
## Summary
This is the initial progress of a possible _future_ replacement of the Prompt interface. The idea is that it allows for an arbitrary number of form inputs each with their own individual form validation. There have been recently discovered issues with form validation that this addresses.

Note that this work is still in progress and should be treated as something to build from, but it is currently needed to resolve client issues as soon as possible.

### Supported Inputs
Currently only two types of inputs are supported (with their own subtypes); Dates and Text. Date inputs support all the existing date formats, includes text masking, min/max validation, etc. Text can either be numeric or arbitrary and can be validated with a number of parameters including minimum length, maximum length, and regular expression pattern.

Text inputs also support scanning directly to the input for both image scanners and traditional scanning. Scanning can also automatically submit the form after validation takes place.

### Future Notes
Currently functionality is limited to two input types but is built around supporting other types such has boolean toggles, drop down selections, etc. These should be added in the future.

Additionally there are some issues with existing input controls working with the form correctly and I couldn't figure out exactly why in the timeframe I had. It would be ideal to investigate this further and use common controls.

### Screenshots
Screen Shot 2021-12-22 at 6 52 51 PM
Untouched Form

Screen Shot 2021-12-22 at 6 53 03 PM
Demonstrating automatic placeholder text for date entry.

Screen Shot 2021-12-22 at 6 53 12 PM
Demonstrating invalid date entry, notice greyed out Next button.

Screen Shot 2021-12-22 at 6 53 27 PM
Valid input immediately after text entry become valid; there is a bug with the existing form where the validation that prevents this behavior from taking place until focus changes first.

Screen Shot 2021-12-22 at 6 53 37 PM
Demonstrates the Image Scanner (Red Box is where the Native Image Scanner would take its place)


chenson42 (approved):

openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/UIMessageType.java
```diff
@@ -6,6 +6,7 @@
     public static final String RETURN = "Return";
     public static final String DIALOG = "Dialog";
     public static final String PROMPT = "Prompt";
+    public static final String PROMPT_PLUS_PLUS = "PromptPlusPlus";
```
eddielegg-le, line 9: This deserves an A+++++++++++++++++++
Yeah we just watched Christmas Story

openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/prompt/TextPromptInput.java
```diff
@@ -0,0 +1,44 @@
...
+@Getter
+@Setter
+public class TextPromptInput extends BaseInputPrompt<String> {
+    private TextPromptInputKind kind;
+    private Integer minimumLength;
+    private Integer maximumLength;
+    private String pattern;
+    private boolean allowBarcodeScanEntry;
+    private String placeholder;
+
```
eddielegg-le, line 16: Think maybe a required flag would be good here too?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/prompt2/form.interface.ts
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * Action that is invoked instead of the `submitAction` if `submitOnScan` is 
+   * enabled. Using this method will provide the raw scan information via.
+   * the action instead of validating with the form.
+   */
+  submitOnScanAction?: string;
+}
+
+export type DynamicFormControlType =
+  //'group'
```
eddielegg-le, line 25: Looks like this guy is getting referenced still

chadhav-ae (dismissed):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/prompt2/form.interface.ts
```diff
@@ -0,0 +1,75 @@
...
+  MMDDYY = 'MMDDYY',
+  MMDDYYYY = 'MMDDYYYY',
+  DDMMYY = 'DDMMYY',
+  DDMMYYYY = 'DDMMYYYY'
+}
+
+export interface DynamicDateInput extends BaseInputControl<Date> {
+  readonly format: DateFormatKind;
+  readonly min?: Date;
+  readonly max?: Date;
```
jchirico8, line 68: Anton had put in changes that I believe I used for date of birth where min was relative. ie we'd say we accept no one more than 100 years old so we could set the min to 100 years and it slides with time instead of picking an arbitrary minimum date or calculating it individually for fields

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/prompt2/prompt2-screen-dialog.component.html
```diff
@@ -0,0 +1,62 @@
...
+    <form class="prompt-content" [formGroup]="rootFormGroup" (submit)="onFormSubmit()">
+        <app-content-card>
+            <ng-container *ngFor="let control of screen.items">
+                <div class="item-group" [ngSwitch]="control.type">
+                    <div *ngSwitchCase="'text'" class="flex col">
+                        <mat-form-field responsive-class>
+                            <mat-label responsive-class>{{ control.label }}</mat-label>
+                            <input matInput responsive-class [type]="control.kind === 'Number' ? 'tel' : 'text'"
+                                [formControlName]="control.id" [attr.minlength]="control.maximumLength"
+                                [attr.maxlength]="control.minimumLength" [placeholder]="control.placeholder"
```
jchirico8, line 13: min=max and max=min?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/prompt2/prompt2-screen-dialog.component.html
```diff
@@ -0,0 +1,62 @@
+<div *ngIf="rootFormGroup && screen" class="prompt-outer">
```
jchirico8, line 1: It would be nice if we could get separate components for these field types. It's fine now, but if this grows with more field types or doubles to add a whole mobile version, it'll be unreadable.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/prompt2/prompt2-screen-dialog.component.html
```diff
@@ -0,0 +1,62 @@
...
+
+                        <app-image-scanner
+                            *ngIf="controlSupportsBarcodeScanInput(control) && activeScanInput == control"
+                            (scan)="acceptScanInputForControl(control, $event)">
+                        </app-image-scanner>
+                    </div>
+
+                    <mat-form-field *ngSwitchCase="'date'" responsive-class>
+                        <mat-label responsive-class>{{ control.label }}</mat-label>
+                        <input responsive-class type="tel" matInput type="text" [formControlName]="control.id"
```
jchirico8, line 37: Do you want two type attributes?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/prompt2/form.interface.ts
kdelorey: I moved them all out into a `todo` comment so future me is aware of the intention.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/prompt2/prompt2-screen-dialog.component.html
kdelorey: 🤦

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/prompt2/prompt2-screen-dialog.component.html
kdelorey: Yeah that the idea for the future. There were just some weirdness with the Angular forms that I had kind of resolved but then it messed with the visual indications of invalid input. I ended up giving up trying to figure it out for now and pressed on.

Definitely want this for the future.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/prompt2/prompt2-screen-dialog.component.html
kdelorey: 🤦

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/prompt2/form.interface.ts
kdelorey: Seems like a server side issue behavior. In the use case for this `min` needs to be `now` as its an expiration date that must be some time in the future. I can't anticipate the usages of this generic entry, I need some help from the state posting the screen.

Also for DoB there are people > 100yo 😄

openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/prompt/TextPromptInput.java
kdelorey: `required` is handled by the base class. Its passed through in the constructors.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/prompt2/form.interface.ts
jchirico8: For future only dates you'd set min to 0 meaning the minimum date is `now - 0`. Dates required to be in the past would be max=0 so max date is `now + 0`

kdelorey: Yeah, I'm mostly interested why the client should arbitrarily figure out a range? It seems like the state should make a determination as to what the range should be since the client can't anticipate its usage. Maybe it's in a use case where its +/- 1 yr, - 5yr, + 100yr, etc.

openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/UIMessageType.java
kdelorey: a-christmas-story

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/prompt2/form.interface.ts
jchirico8: You're right. I don't have a clear enough picture of all the logic that goes into building this form so I confused myself. This would be something built on the server and you haven't touched

jchirico8 (approved):

PR #1829, opened 2021-12-30
Bugfix/scan control char stripping

shiljo-litmus7:
Issues Fixed
[PDPOS-6785](https://petcoalm.atlassian.net/browse/PDPOS-6785)

On mobile devices, when scan new pals ID on Add Loyalty Customer screen for Loyalty number field device displays "input is invalid"

Summary
Mobile devices uses wedge scanner which emulates keystrokes. In device's scanner profile, we would have set the start control character for application to detect scans. This start control char needs to be stripped out from Loyalty Number Field


jrmihalick (approved):

jrmihalick:
Nice job considering the lack of good options we had to implement this.

PR #1837, opened 2022-01-04
Account Lookup Display Adjustment

coohanlonCode:
### Summary
Implements the UI aspect during a return to the SaleState to display dynamic text describing that an account lookup is occurring on the CustomerDisplay instead of showing an undesired State/wait window.


jchirico8:
Could you add a screenshot?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/sale-total-panel/sale-total-panel.component.html
```diff
@@ -117,9 +117,16 @@ <h4 class="membership-rewards-title" responsive-class>{{screenData.rewardsLabel}
...
-                            {{screenData.noPromotionsLabel}}
-                        </div>
+                        <ng-container *ngIf="!screenData.loyaltyRewards?.length" class="membership-no-rewards">
+                            <div>
+                                {{screenData.noPromotionsLabel}}
+                            </div>
+                        </ng-container>
+                        <ng-container *ngIf="screenData.cancelAcctLookupProcessVisible">
+                            <div>
+                                <span>Click to Cancel Process</span>
```
chenson42, line 127: This should not be hard coded.  It should come from a resource bundle from the server

chenson42: Wondering if this property should be more generic?  cancelAcctLookupProcessVisible

jchirico8:
I don't think this change is necessary. It looks like you were matching the mock ups too closely. We already use this to notify the cashier of what the customer is doing and you'd just want to send an event to change the message 
image

coohanlonCode:
> I don't think this change is necessary. It looks like you were matching the mock ups too closely. We already use this to notify the cashier of what the customer is doing and you'd just want to send an event to change the message image

I have not seen this screen before and neither has the tester logging the bug. The state that we were observing was the popup window displaying as per below.
image

Which branch and OS were you able to create that screen?

jchirico8:
That's the add customer flow when you pass the form to the customer display

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/sale-total-panel/sale-total-panel.component.html
coohanlonCode: Hard-coded text has been been converted via @ e4a27f5

coohanlonCode:
> That's the add customer flow when you pass the form to the customer display

The [JIRA bug PCOM-5989](https://jira.ae.com/browse/PCOM-5989) that this PR is addressing is only related to how the SaleState screen is presented once an existing and valid loyalty ID is successfully provided. 

Adding a customer through the loyalty signup process is not in scope for this fix to the SaleState screen presentation.

jchirico8:
It's the same concept and it should work the same. `CustomerDisplayLoyaltySignupEventHandler` has action handlers that call its `updateInProgress` helper with a message pulled from i18n. A lot of it is named after loyalty signup but it's very reusable

chenson42:
I'm having a hard time following this.  Did @coohanlonCode make the adjustments you recommended @jchirico8 ?

jchirico8:
@chenson42 no this is still using plain text on the customer tile instead of the notification bell tile

coohanlonCode:
Do I need to modify the flows so that the Account Lookup is the same as the Loyalty Signup?

Are they supposed to be the reused Loyalty Signup state rather than the modification to the SaleState as shown in the Jira ticket and product description?

jchirico8:
@blubaugha you did all of the notification stuff. Most of it refers to loyalty sign up but I imagine some of your code can be reused for Colin's feature. How difficult would it be to trigger the `isLoyaltySignupInProgressOnCustomerDisplay$` for AECC lookup and have all new messaging?

blubaugha:
> @blubaugha you did all of the notification stuff. Most of it refers to loyalty sign up but I imagine some of your code can be reused for Colin's feature. How difficult would it be to trigger the `isLoyaltySignupInProgressOnCustomerDisplay$` for AECC lookup and have all new messaging?

@jchirico8 @chenson42 
I'm having a hard time following this too. It's not clear to me what problem is being solved and why we need this change. Can you summarize for my simple brain @jchirico8 ?

jchirico8:
When looking up an AECC account, you can have the customer enter their information into the search. This is the notification bell to keep the cashier updated on what the customer is doing

blubaugha:
> When looking up an AECC account, you can have the customer enter their information into the search. This is the notification bell to keep the cashier updated on what the customer is doing

@jchirico8 thanks, my simple brain has processed your summary.

@coohanlonCode 
You can add a new event type and publish it from the server so it shows on the cashier's screen.

Here's how to do that:

1. Add a new signup event type for this scenario
**commerce-sales-api/src/main/java/org/jumpmind/pos/sales/event/LoyaltySignupUpdateEventType.java**
```
public static final LoyaltySignupUpdateEventType CUSTOMER_PRIVATE_LABEL_CARD_SIGNUP = new LoyaltySignupUpdateEventType("customerPrivateLabelCardSignup");
```

2. Publish the event where needed
```
// If you're publishing it from a state there's a helper method
publish(new LoyaltySignupUpdateEvent(LoyaltySignupUpdateEventType.CUSTOMER_PRIVATE_LABEL_CARD_SIGNUP, customer));

// If you're doing it outside of a state
@Autowired
EventPublisher eventPublisher;

LoyaltySignupUpdateEvent event = new LoyaltySignupUpdateEvent(LoyaltySignupUpdateEventType.CUSTOMER_PRIVATE_LABEL_CARD_SIGNUP, customer);
event.setDeviceId(appContext.getDeviceId());
event.setPairedDeviceId(appContext.getPairedDeviceId());
event.setAppId(appContext.getAppId());

eventPublisher.publish(event);
```

3. Add an i18n message for **Commerce** (just so it's in the base code for use outside of AEO). _NOTE_: The `customerPrivateLabelCardSignup` string used for the event is the last part of the key.
**aeo-server/src/main/resources/i18n/sale-override_en.properties**
```
screen.loyaltyRewards.message.customerPrivateLabelCardSignup=Customer is signing up for private label card
```

4. Add an i18n message specific to AEO
**aeo-server/src/main/resources/i18n/sale-override_en.properties**
```
screen.loyaltyRewards.message.customerPrivateLabelCardSignup=Customer is signing up for AECC
```

kdelorey:
Similar to your commerce PR, this PR now has a bunch of diffs for unrelated things that I think are already merge. I'm not sure what's going on, but its very noise to review.

PR #1853, opened 2022-01-07
PCOM-6286 Minimum Date Validation Failed

chadhav-ae:
### Issues Fixed
Minimum Date Validation Failed on same day.


kdelorey (approved):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/validators/custom-date-validators.ts
```diff
@@ -9,6 +9,7 @@ export class CustomDateValidator {
                 const inputDate = new Date(field.value);
                 if (!isNaN(inputDate.getTime())) {
                     const minDate = this._resolveTime(minTime);
+                    minDate.setDate(minDate.getDate() - 1);
```
kdelorey, line 12: This should be done at the server not here. The server sets the minDate as the current date and it needs to be "- 1". This change here affects all date validation now.

kdelorey (changes requested):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/validators/custom-date-validators.ts
kdelorey: Change needs to happen in `TenderShoppingPassState.java` line `220`

chadhav-ae: But does we know all date are not failing on same day. I updated the server first than found all date should be failing this check if we use min date as same date.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/validators/custom-date-validators.ts
```diff
@@ -9,6 +9,7 @@ export class CustomDateValidator {
...
                     if (inputDate < minDate) {
```
jchirico8, line 13: Couldn't you just do this?
```suggestion
                    if (inputDate <= minDate) {
```

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/validators/custom-date-validators.ts
kdelorey: You should also keep in mind that `Date` objects also include `Time`. I think what you really want to do is construct a date of "Today" at 00:00:00

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/validators/custom-date-validators.ts
chadhav-ae: This is not working as well as inputDate is having a time and min Date is having time which is different. So the date comparing is showing a different result.

jchirico8: That makes sense. I think Kevin's idea to set the time on min date would work best

PR #1858, opened 2022-01-07
PCOM-6298 Add validation to make shopping pass text field readonly

chadhav-ae:
### Issues Fixed
Adding a validation to make shopping pass text field as readonly.


openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/prompt2/form.interface.ts
```diff
@@ -47,6 +47,7 @@ export interface DynamicTextInput extends BaseInputControl<string> {
   readonly pattern?: string;
   readonly allowBarcodeScanEntry?: boolean;
   readonly placeholder?: string;
+  readonly readOnly?: boolean;
```
kdelorey, line 50: I could see this being promoted to the `BaseInputControl<>` class and everything handles it, but that probably needs to be thought out some more so I'm merely noting it for now.

kdelorey (approved):

PR #1864, opened 2022-01-07
Fix typo in platform-plugins.startup-task.ts causing scanners to be removed from the SCANNERS array

### LINKED ISSUE #1863: JPOS scans are not received by the server
For configurations that are using a JPOS scanner, which sends the scan from the device running the commerce server to the client, the scan data is not being received back by the server.

PDPOS-7149
[3.0][Scanner] Scanner not responding when scanned on Sale and Return screen

**Expected behavior**
The scan should be received on the server and processed like it used to prior to recent changes on master.

jrmihalick:
- This was causing valid/active scanners (server scanner and console scanner) to be removed from the array of scanners
- Also add support for single argument `console.scan('1234567')` like we used to have

### Issues Fixed
#1863


chenson42 (approved):

kdelorey (approved):
Whoops 😅

PR #1878, opened 2022-01-11
Keyboard always pops on each screen on iOS

chenson42:
Keyboard always pops on each screen on iOS.

@mmichalek What did I break by removing your fix??


mmichalek:
It's a workaround for a visual dark circle around the upper left icon, which turns out to be tricky to get rid of. I'll see if I can find another workaround.

mmichalek (approved):

PR #1916, opened 2022-01-17
Document gets reloaded once on failed personalization request

### LINKED ISSUE #1915: Auto Personalization requests fail in stores running in electron until page is reloaded
Specifically in stores running in electron the auto personalization request will fail simply from being timed out from the server. It seems like it doesn't even to make an attempt to establish a connection to the server. If you use the dev tools to then refresh the page (don't close the electron app) the request will then succeed. An attempt to solve this by retrying the request multiple times did not resolve the issue.

kdelorey:
Resolves #1915


jrmihalick (approved):
I just realized this was on 2.6 and not master. Approving in case you need it to move forward.  Hopefully this won't need to come to master?

kdelorey:
@jrmihalick no idea, `master`'s whole startup sequence is completely different so I don't intend to. If AEO started building from `master` then we might have an issue.

But yeah I agree with your feelings, like I was telling @chenson42:
> kinda feels dirty

To be honest I can't confirm this will even "fix" it, but it fixes it when I triggered it by hand, so we'll see.

PR #1920, opened 2022-01-18
Enhance startup logging, add overload of SessionService.sendMessage()

### LINKED ISSUE #1919: Enhancements to client startup task logging
**Is your feature request related to a problem? Please describe.**
- After compiling the client app in production mode, we lose names of startup tasks in the startup task logging.  E.g., logging like this occurs:

Screen Shot 2022-01-13 at 6 13 52 PM

- We also are not logging any messages from the platform plugins as they initialize
- We also need an overload of the `ProxyService.sendMessage()` to allow for passing in a device ID.  Messages are not making it to the NCR plugin when running on Android because the messages are being sent to the payment device deviceId and not the mobile device deviceId.

jrmihalick:
### Issues Fixed
#1919


kdelorey (approved):
lgtm

PR #1930, opened 2022-01-21
Revert auto-personalization hack in favor of much longer attempt

### LINKED ISSUE #1915: Auto Personalization requests fail in stores running in electron until page is reloaded

kdelorey:
Reverts #1916 
Resolves #1915


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/personalization/personalization.service.ts
```diff
@@ -133,44 +133,23 @@ export class PersonalizationService {
         Observable<AutoPersonalizationParametersResponse> {
         let url = this.sslEnabled$.getValue() ? 'https://' : 'http://';
         url += `${config.ipv4Addresses[0]}:${config.port}/${config.txtRecord.path}`;
+
+        const requestTimeout = 2000;
+        const retryCount = CONFIGURATION.autoPersonalizationRequestTimeoutMillis / requestTimeout;
```
jchirico8, line 138: we retry 2.5 times?

jchirico8 (approved):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/personalization/personalization.service.ts
kdelorey: 50000 / 2000 = 25

jchirico8: I suck at reading large numbers without comma separation

Issue #1931, opened 2022-01-21
Options List not defaulting to focusing the first option


brandon-lewis:
**Describe the bug**
[Reported by Petco]
When viewing any options-list component, the first option in the list is not default focused. This happens on every options list.

**To Reproduce**
1. Log in and access the sale screen
2. Select "Gift Cards" in the left hamburger nav

**Expected behavior**
When viewing any options-list component, the first option in the list should be default focused.

PR #1932, opened 2022-01-21
New config for auto-focus first options list option

brandon-lewis:
### Issues Fixed
https://github.com/JumpMind/openpos-framework/issues/1931

Added a new `autoFocusFirstOptionsListOption` CONFIGURATION key. The key value defaults to false on the openpos side, but the configurable value will be fed to the openpos client application from the server for determining whether to auto-focus the first option in the options list.


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/options-list/options-list.component.ts
```diff
@@ -39,7 +40,7 @@ export class OptionsListComponent extends ScreenPartComponent<OptionsListInterfa
 
     isMobile: Observable<boolean>;
 
-    isFirstElementFocused = false;
+    autoFocusFirstOption;
```
kdelorey, line 43: can you not just combine this with what's in the ctor?

```ts
autoFocusFirstOption = CONFIGURATION. autoFocusFirstOptionsListOption;
```
If not can we explicitly type this so it's not as confusing?

```suggestion
    autoFocusFirstOption: boolean;
```

kdelorey (approved):

Issue #1947, opened 2022-01-27
Not All Dialogs Close when Pressing Escape


blubaugha:
### Summary
Users expect dialogs to close when pressing the `Escape` key. Not all dialogs/prompts have a `Close` action set, so this causes usability issues for users because some dialogs close when pressing the `Escape` key and others do not.

All dialogs should have consistent closing behavior when pressing the `Escape` key.

Instead of updating all of the dialogs/prompts that are missing a `Close` action for the `Escape` key, dialogs should have default behavior to close when pressing the `Escape` key to ensure a consistent user experience.

PR #1933, opened 2022-01-21
Remove Old Keybinding Code and Bug Fixes

blubaugha:
# Don't be scared of all of the modified files! It's mostly deleted code!

### Issues Fixed
Keybindings Need Rewritten
[1874](https://github.com/JumpMind/openpos-framework/issues/1874)

### Summary
This removes all of the old keybinding code and adds bug fixes and more tests.

### Bug Fixes
* `PopTartComponent` didn't implement new keybindings
* Searching for items, on the **Sale** screen, executed the action twice: once with a payload and once without
* Kebab menu closed after every keypress
* `CustomerSearchResultDialogComponent` didn't pass the selected index in the action payload when using the `Enter` key

PR #1948, opened 2022-01-27
Keybinding Dialog Behavior - Close on Escape by Default

blubaugha:
### Issues Fixed
[5341](https://github.com/JumpMind/commerce/issues/5341)
[1947](https://github.com/JumpMind/openpos-framework/issues/1947)

### Notes
I added a new singleton service, called `KeybindingDialogService`, for handling keybinding behavior specific to dialogs. It will automatically send a `Close` action to the server, when pressing the `Escape` key, if the there is no keybinding for the `Escape` key in the active keybinding zone.

If we need more dialog specific behavior for keybindings in the future, we can add it to this service.

### Screenshots
Close when there's no keybinding for the `Escape` key

dialog-no-close-action

Don't handle closing when a keybinding exists for the `Escape` key

dialog-with-close-action


blubaugha:
It's strange that Sonar code coverage somewhat worked (results were there but not accurate) in my PR, but doesn't work in this PR and nothing has changed in the `karma.conf.js`.


https://github.com/JumpMind/openpos-framework/pull/1933
image

kdelorey (approved):

PR #1953, opened 2022-01-28
Restricted JSON sanitize to a single level of depth.

### LINKED ISSUE #1952: Logging sanitizer can overflow causing crash
It is possible to cause the JSON sanitizer to overload an crash. Easy reproduction:
1. Enable `server-logging`
1. Run Commerce Demo Store
1. Press Sale
1. Login
1. Logout
1. Boom!

kdelorey:
Resolves #1952

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/logging/console-scraper.service.ts
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/logging/console-scraper.service.ts
@@ -112,7 +112,7 @@ export class ConsoleScraper {
     private _makeCircularRefReplacer = () => {
         return (key: string, value: any) => {
             if (!!key && typeof value === 'object' && value !== null) {
-                return "[Object object]";
+                return '[Object object]';
             }
 
             return value;
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/logging/console-scraper.service.ts
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/logging/console-scraper.service.ts
@@ -110,15 +110,9 @@ export class ConsoleScraper {
     }
 
     private _makeCircularRefReplacer = () => {
-        const seen = new WeakSet();
-
-        return (_key: string, value: any) => {
-            if (typeof value === 'object' && value !== null) {
-                if (seen.has(value)) {
-                    return;
-                }
-
-                seen.add(value);
+        return (key: string, value: any) => {
+            if (!!key && typeof value === 'object' && value !== null) {
+                return "[Object object]";
             }
 
             return value;


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/logging/console-scraper.service.ts
```diff
@@ -110,15 +110,9 @@ export class ConsoleScraper {
...
-        const seen = new WeakSet();
-
-        return (_key: string, value: any) => {
-            if (typeof value === 'object' && value !== null) {
-                if (seen.has(value)) {
-                    return;
-                }
-
-                seen.add(value);
+        return (key: string, value: any) => {
```
jrmihalick, line 113: This looks like it will never print objects at the first level either and will just always print `[Object object]`?  I am wondering why your prior logic did not work?  Was it because there was something wrong with `typeof value === 'object' && value !== null`?

kdelorey: `key` is `undefined` at the first level. The condition has changed to `!!key && value is object && value != null`

kdelorey: Other than that, there is no other way to control depth short of making an entirely new json stringify, which has been done but its a bit nasty.

jchirico8 (approved):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/logging/console-scraper.service.ts
kdelorey: The problem is that something (I couldn't figure out what during the crash) is serializing too big of an object and causing an overflow; this was the resolution to it.

jrmihalick: I wonder if this is only happening because of @blubaugha logging the zones in the keybinding classes and this problem will go away once he removes that code.

kdelorey: Has the issue been resolved? Do you want me to close this?

chenson42: I was gonna ask the same, @jrmihalick ?

jrmihalick: Yes, we are fine without this change.  @kdelorey This can be closed.

PR #1975, opened 2022-02-08
Verify Id prompt for new PLCC signup - PLCCApplyCheckIDState- added secondary button for dynamic form

devallur:
Verify Id prompt for new PLCC signup - PLCCApplyCheckIDState
Add secondary  buttons for the valid ID screen using dynamic form . secondary button will be displayed only when the button is set with actionItem value.

Screen Shot 2022-02-08 at 2 16 21 AM

commerce PR -https://github.com/JumpMind/commerce/pull/5506


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/dynamic-form-part/dynamic-form-part.component.html
```diff
@@ -16,6 +16,10 @@
                 <span>{{button.label}}</span>
                 <app-icon *ngIf="button.icon" [iconName]="button.icon"></app-icon>
             </app-secondary-button>
+            <app-secondary-button *ngFor="let button of secondaryButtons" (click)="doAction(button)">
```
jchirico8, line 19: can you just change the `let button of buttons` to be secondaryButtons? I don't see buttons on the dynamic form interface or uimessage so I don't know why they're in here

devallur:
@jchirico8  As per the review comment , i have made changes to replace buttons with secondaryButtons as buttons is not defined in uimessage or interface and is not in use

jchirico8:
> @jchirico8 As per the review comment , i have made changes to replace buttons with secondaryButtons as buttons is not defined in uimessage or interface and is not in use

This is my fault. I deleted my comment instead of saying I was wrong so you still saw it in the email. I was looking at the wrong dynamic form component. The buttons are used and are constructed in the angular component using button typed form elements. I'm not sure what exactly might be using that, but you should add it back

devallur:
> > @jchirico8 As per the review comment , i have made changes to replace buttons with secondaryButtons as buttons is not defined in uimessage or interface and is not in use
> 
> This is my fault. I deleted my comment instead of saying I was wrong so you still saw it in the email. I was looking at the wrong dynamic form component. The buttons are used and are constructed in the angular component using button typed form elements. I'm not sure what exactly might be using that, but you should add it back



> > @jchirico8 As per the review comment , i have made changes to replace buttons with secondaryButtons as buttons is not defined in uimessage or interface and is not in use
> 
> This is my fault. I deleted my comment instead of saying I was wrong so you still saw it in the email. I was looking at the wrong dynamic form component. The buttons are used and are constructed in the angular component using button typed form elements. I'm not sure what exactly might be using that, but you should add it back

Thanks Joe . i have reverted the changes done earlier .

VodianytskyiVI (approved):

PR #1994, opened 2022-02-16
Support for Zebra Printer Status Code Handling / Cover Open Check

### LINKED ISSUE #1993: Zebra Printer Is Not Producing Error Statuses (Cover Open Specifically)
**Describe the bug**
Current implementation is returning false for any status check on Zebra Printers. This is due to the Zebra printers requiring ZPL for status retrieval, unimplemented currently.

**Expected behavior**
Ability to retrieve status codes from the Zebra printers.

liambell-le:
### Issues Fixed
#1993


openpos-peripheral/src/main/java/org/jumpmind/pos/print/ZebraPrinter.java
```diff
@@ -227,8 +234,42 @@ public PeripheralConnection getPeripheralConnection() {
     }
 
     @Override
+    @SneakyThrows
```
chenson42, line 237: @jrmihalick Please review.  I think we might need to do something here in order in order for the blue tooth approach to work.

@mmichalek we had trouble getting the pure protocol to work.  We ended up pulling in a zebra sdk jar file.

chenson42 (approved):

jrmihalick (approved):

PR #2018, opened 2022-02-22
Set a timeout on stream reads from the zpl port connection

### LINKED ISSUE #2013: Timeout required for Zebra Printer Status Connection
**Describe the bug**
The default timeout of the ZQ320 is around 6 seconds. Change the status timeout to match that of the SocketConnectionFactory timeouts.

liambell-le:
### Issues Fixed
#2013


openpos-peripheral/src/main/java/org/jumpmind/pos/print/ZebraPrinter.java
```diff
@@ -243,7 +243,7 @@ public int readPrinterStatus() {
             try {
                 connection.open();
                 com.zebra.sdk.printer.ZebraPrinter printer = ZebraPrinterFactory.getInstance(connection);
-
+                printer.getConnection().setMaxTimeoutForRead(2500);
```
chenson42, line 246: look in the socket connection factory.  there is an example of getting the timeout from settings

Issue #2031, opened 2022-03-01
Support server-driven locale configs that also drive client behavior


brandon-lewis:
**Is your feature request related to a problem? Please describe.**
Other tickets and issues are in progress for i18n and l10n. Currently, there is some content that is not i18n-able through device configs. It would be nice to be able to configure a businessUnit locale that would drive the entire commerce and openpos experience to be internationalized. This is not a bug, but something the product should be able to do.

PR #2032, opened 2022-03-01
Feature/es i18n

brandon-lewis:
### Solution Details
Updated the dynamic-date-form-field component to pull in a configured locale from the LocaleService. The component uses angular components and tooling that supports i18n out-of-the-box, so adding the adapter to set the locale retrieved from the LocaleService will automatically update the client to use the appropriate angular-provided i18n for the calendar widget.

There was also an issue with the content providers in that they were fine until we wanted to have multiple potential variations on configured `openpos.ui.content.file-system.providerProperties` (previously just brandId and deviceType, but also now to include locale). The update to the AbstractFileContentProvider is to provide all permutations of the configured provider properties. Before this change, a configured provider props of "a, b, c" would return the following filepaths for base "myContentDir":
- myContentDir/a/b/c
- myContentDir/a/b
- myContentDir/a
- myContentDir/

With the changes here, we will get all possible permutations for the client to decide how to configure their provider props without also requiring unnecessarily deeply nested dirs. The same content provider props for the same base will now return a structure that looks more like the following:
- myContentDir/a/b/c
- myContentDir/a/c/b
- myContentDir/b/a/c
- myContentDir/b/c/a
- myContentDir/c/a/b
- myContentDir/c/b/a
- myContentDir/a/b
- myContentDir/a/c
- myContentDir/b/a
- myContentDir/b/c
- myContentDir/c/a
- myContentDir/c/b
- myContentDir/a
- myContentDir/b
- myContentDir/c
- myContentDir/

Note that it follows a "deepest first" strategy for prioritization. The content provider system will search through that list top-to-bottom for content to exist in the current path, and will continue on until it finds content or all paths have been searched.

The final change is to the LocaleChangedMessage to _actually_ pass a configured locale to the client. The changes here precede a commerce PR.

### Issues Fixed
https://github.com/JumpMind/openpos-framework/issues/2031


openpos-flow/src/main/java/org/jumpmind/pos/core/content/AbstractFileContentProvider.java
```diff
@@ -98,26 +100,52 @@ public String getMostSpecificContent(String deviceId, String key, String baseCon
         return Collections.emptyList();
     }
 
-    protected List<String> getPossibleContentDirs(String key) {
-        List<String> possibleContentDirs = new ArrayList<>();
+    protected List<String> getAllPossibleContentDirPermutations(String key) {
```
jrmihalick, line 103: Should we use caching here?

openpos-flow/src/main/java/org/jumpmind/pos/core/content/AbstractFileContentProvider.java
```diff
@@ -98,26 +100,52 @@ public String getMostSpecificContent(String deviceId, String key, String baseCon
...
-                    if (personalizationProperties.containsKey(property)) {
-                        builder.append("/").append(personalizationProperties.get(property));
+            List<String> additions = new ArrayList<>();
+            permutations.forEach(s -> {
+                for(int x = 1; x < s.length(); x++) {
+                    if(s.charAt(x) == '/' && !additions.contains(s.substring(0, x) + "/")) {
+                        additions.add(s.substring(0,x) + "/");
                     }
                 }
+            });
```
kdelorey, line 116: Am I mistaken in thinking this is effectively a split?

Maybe something like this (you may have to play with it) will give you the same effect with fewer iterations of each string separation. You could actually probably simplify this further.

I was also curious about the inclusion of `additions.contains(s.substring(0, x) + "/")` . This kind if ruins the depth sort if your directories are of equal names. Anyways I figured I just mention it.

```java
            return permutations.stream()
                .map(s -> s.split('/').stream().distinct().map(z -> key + z).collect(Collectors.toList()))
                .sorted(Comparator.comparingInt(List::size))
                .map(s -> String.join("/", s))
                .collect(Collectors.toList());
```

openpos-flow/src/main/java/org/jumpmind/pos/core/content/AbstractFileContentProvider.java
kdelorey: Probably a good idea. Figuring this information out can basically `O(n!)`, caching it would be a nice way of making `O(1)`.

openpos-flow/src/main/java/org/jumpmind/pos/core/content/AbstractFileContentProvider.java
```diff
@@ -98,26 +100,52 @@ public String getMostSpecificContent(String deviceId, String key, String baseCon
...
-                possibleContentDirs.add(builder.toString());
+            additions.sort((s1, s2) -> s2.split("/").length - s1.split("/").length);
+            permutations.addAll(additions);
+            permutations = permutations.stream().map(s -> key + s).collect(Collectors.toList());
+        }
+        return permutations;
+    }
+
+    private void getPermutations(List<String> values, String[] array, int pos){
+        if(pos >= array.length - 1){
```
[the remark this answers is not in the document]
brandon-lewis: Good catch, must easier to read. I was dealing with some bugfixes yesterday and this was one of those "I think this change might fix it" and I overlooked that clarity.

openpos-flow/src/main/java/org/jumpmind/pos/core/content/AbstractFileContentProvider.java
```diff
@@ -98,26 +100,52 @@ public String getMostSpecificContent(String deviceId, String key, String baseCon
...
 
-                possibleContentDirs.add(builder.toString());
+            additions.sort((s1, s2) -> s2.split("/").length - s1.split("/").length);
+            permutations.addAll(additions);
+            permutations = permutations.stream().map(s -> key + s).collect(Collectors.toList());
+        }
+        return permutations;
+    }
+
+    private void getPermutations(List<String> values, String[] array, int pos){
```
jchirico8, line 125: Could we use better variable names? I had to scroll back and forth to try and straighten out what array was and pos makes me think of point-of-sale
```suggestion
    private void getPermutations(List<String> values, String[] provderProperties, int position){
```

openpos-flow/src/main/java/org/jumpmind/pos/core/content/AbstractFileContentProvider.java
```diff
@@ -109,42 +107,47 @@ public String getMostSpecificContent(String deviceId, String key, String baseCon
...
-        if(pos >= array.length - 1){
+    private void getPermutations(List<String> values, String[] shiftablePermutationValues, int pos){
+        if(pos >= shiftablePermutationValues.length - 1){
             StringBuilder sb = new StringBuilder();
-            for(int i = 0; i < array.length - 1; i++){
-                String property = array[i];
-                if (personalizationProperties.containsKey(property) && !values.contains("/" + personalizationProperties.get(property))) {
+            for(int i = 0; i < shiftablePermutationValues.length - 1; i++){
+                String property = shiftablePermutationValues[i];
+                // && !values.contains("/" + personalizationProperties.get(property))
```
kdelorey, line 132: Can this be removed?

brandon-lewis: @kdelorey Oh good catch on the comment, thank you

jchirico8 (approved):

PR #2048, opened 2022-03-03
Bugfix/hidden element focus breaks autofocusfirstoption config

stojankz:
The focus on a hidden element on bacon strip causes options list configuration not to work when the options list is part of a screen rather than dialog because the bacon strip has a hidden element with focus. For example, on the tender screen, the first option is not focused even though the autoFocusFirstOptionsListOption config is set to 'true'. Remove hidden element from bacon-strip.

Before:
autofocus-bug

After:
autofocus-bugfix


jrmihalick:
Is there any way that the bacon strip can bypass its focus on the hidden button if there is any other component in the dom that has cdkFocusInitial?  @kdelorey Any alternate ideas here?

stojankz:
> Is there any way that the bacon strip can bypass its focus on the hidden button if there is any other component in the dom that has cdkFocusInitial? @kdelorey Any alternate ideas here?

Yeah I didn't think that the bacon strip has a way to check for child elements in the screen so this was the best workaround I could think of. Overall, I'm not entirely sure whether the hidden element in the bacon strip is good either. I think @mmichalek added that for NRF to get rid of the focus that was on the icon on the top left but this is definitely an unintended consequence of this hidden element.

kdelorey:
> Is there any way that the bacon strip can bypass its focus on the hidden button if there is any other component in the dom that has cdkFocusInitial? @kdelorey Any alternate ideas here?

I think the ideal situation would be to not have that hidden button at all, but it is solving a problem I suppose. Maybe figuring out how to get around that initial problem that @stojankz mentioned would be the long term fix. I remember @mmichalek struggling with it though. Trying to solve the problem as is I don't have any good solutions off the top of my head, I'll need to think on it some more.

stojankz:
Added another solution that handles setting the focus on the second element that has cdkFocusInitial after the hidden element in the bacon-strip. For the record, this is also a hacky solution, but I suppose it is a little more dynamic

jrmihalick (approved):

PR #2057, opened 2022-03-07
fix 'Complete Return button keybind is not enabled after item added #2056'

### LINKED ISSUE #2056: Complete Return button keybind is not enabled after item added
**Describe the bug**
After adding an item to the return screen, the 'Enter' button keybind does not send the 'Complete Return' button action. When debugged, it appeared that the action tied to the keybind still has the enabled field set to false and did not update it with the new screen data. If you leave the Return screen and come back, the Enter keybind works.

**To Reproduce**
1. Go to 'Returns'
2. Click on Scan or Search and enter item '870480'
3. Click 'Enter' button on keyboard
4. See that it does not take you back to Sale

**Expected behavior**
Enter keybind should take you to Sale screen after item is added and 'Complete Return' button is enabled

**Screenshots**
return-enter-click

stojankz:
### Issues Fixed
Closes #2056 

Fix:
return-enter-click-fix


blubaugha (approved):
Nice catch!

jchirico8 (approved):

PR #2067, opened 2022-03-09
Add error for dynamic form field

VodianytskyiVI:
Add error for dynamic form field.
Error is received from backend and displayed as common validation error
Result:
Error


devallur (approved):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/dynamic-form-field/dynamic-form-field.component.ts
```diff
@@ -135,6 +137,14 @@ export class DynamicFormFieldComponent implements OnChanges, OnDestroy, AfterVie
 
     }
   }
+  
+  ngOnInit() {
```
jchirico8, line 141: Don't you have to add `OnInit` to the implements for this to be called?

PR #2071, opened 2022-03-09
fix 'No way to remove filter for device on transaction search when viewed on mobile #2070'

### LINKED ISSUE #2070: No way to remove filter for device on transaction search when viewed on mobile
On desktop view in transaction search, we auto apply a quick filter for the current device on the transaction search. This quick filter can be selected to show transactions across devices for that store. On mobile view however, the filter is not displayed and so there is no way for the user to view transactions for other devices in the store.

**Screenshots**
On Desktop View:
quick-filter

Mobile:
Screen Shot 2022-03-09 at 3 42 48 PM

stojankz:
update transaction search component so that quick filters are visible on mobile view as well as the desktop view 

### Issues Fixed
Closes #2070 

### Screenshot
quick-filter-fix


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/transaction-summary/transaction-summary.component.html
```diff
@@ -99,7 +99,7 @@
         </div>
     </div>
     <div class="summary-action" responsive-class>
-        <app-icon-button *ngIf="transactionSummary.actions"
+        <app-icon-button *ngIf="transactionSummary.actions && (transactionSummary.actions[transactionSummary.actions.length - 1])?.icon"
```
jrmihalick, line 102: Is it possible for transactionSummary.actions to be an empty array?  If so, will angular be unhappy with this?

jchirico8 (approved):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/transaction-summary/transaction-summary.component.html
stojankz: yeah i think that was the error that was occurring in the console. The '?' that I added protects against .icon being called against an undefined element
Screen Shot 2022-03-09 at 4 54 12 PM
.

jrmihalick (approved):

Issue #2074, opened 2022-03-10
Eliminate Duplicate Loyalty/Membership Item Display On Mobile


JoeBrownlee:
**Describe the bug**
For some reason, the loyalty and membership information was being displayed 3 times on mobile clients. Looking at the redundant code, I'm guessing this was from an errant merge. Note that this does **_not_** happen on desktop footprints because a different component is used in that case.

**To Reproduce**
1. Run a mobile footprint. For example, select dimensions like 1000x840. Anything with 840 or less one direction should do it. 
2. Click on 'Sell'.
3. Log in. Cashier will do.
4. Link a customer with loyalty and/or memberships. Both is preferable.
5. Loyalty and membership info is displayed 3 times. See picture below.

**Expected behavior**
Loyalty and membership info should be displayed just once.

**Screenshots**
![image](https://user-images.githubusercontent.com/24355694/157672127-4b2d17f9-1a74-4a55-a9a7-b52c92f5ba19.png)

**Additional context**
Petco reported this issue in [PDPOS-7545](https://petcoalm.atlassian.net/browse/PDPOS-7545).

PR #2075, opened 2022-03-10
Fix Redundant Display Of Loyalty/Membership Info On Mobile

JoeBrownlee:
Petco reported that loyalty/membership info was being displayed 3 times on mobile. This change removes the redundant code that was causing this.

### Issues Fixed
Resolves [#2074](https://github.com/JumpMind/openpos-framework/issues/2074).

Reported by Petco in [PDPOS-7545](https://petcoalm.atlassian.net/browse/PDPOS-7545).


jchirico8 (approved):
Looks like there were some seriously messed up merges in here

PR #2081, opened 2022-03-11
added clone functionality to fix an issue around item caching + getti…

### LINKED ISSUE #2080: Clone item models to prevent object mutation of the shared cached object
Currently ItemModel objects are cached on the FindItem behavior. Since the reference that is returned from the cache is being mutated in place, this impacts all calls that happen afterwards. The specific scenario is the FindLowestPrice behavior modifying the sellingPrice. So if a user goes through that flow and the lowest price is modified to be less than the current price, when the next customer buys that product at the same device it will be stuck on the price found for FindLowestPrice instead of the current price. In order to maintain behavior, but reduce impact this issue requests adding of clone functions/behavior to the ItemModel and supporting model/abstract classes.

r-cochran:
### Issues Fixed
#2080


r-cochran:
Closing, found ObjectUtils.deepClone to use instead.

Issue #2085, opened 2022-03-11
Some modal dialogs unusable on mobile when soft keyboard is showing


brandon-lewis:
**Describe the bug**
Some dialogs are too big for mobile resolutions when the software/OS keyboard is showing.

**To Reproduce**
1. Log in and start a sale.
2. Open the "Search for Customer" dialog to link a customer to the transaction.
3. Set the screen size on the browser to 360x316 (this is the active resolution when the soft keyboard is open on a Zebra TC52 device).
4. Notice that functional parts of the dialog are no longer visible for the user to interact with.

**Expected behavior**
The dialog should be usable (action buttons and inputs should be accessible, dialog should be closeable, etc.).

**Screenshots**
![image](https://user-images.githubusercontent.com/28220454/157976164-fdf8d9f3-ba1e-4f80-9c4b-c0142106b69b.png)

PR #2086, opened 2022-03-11
Fixed unusable modal dialogs for mobile & other minor UI bugs

brandon-lewis:
### Issues Fixed
https://github.com/JumpMind/openpos-framework/issues/2085

### Solution details
There are some screens that still do not recognize "mobile_landscape" as a mobile screen - so when the soft keyboard shows up, it changes the screen resolution to something that is recognized as landscape and does not display the screen correctly.

There was also an issue where dialog buttons were wrapping awkwardly on the customer details dialog, especially on mobile resolutions when the soft keyboard was open. I changed the styling of the dialog action buttons on mobile for a *slightly* better and less awkward display.
image

image

There was also an issue where the "single page application"-ish sub-navigation in the customer details dialog did not work as intended because of overlap between the "close" action and the "back" action. This issue also prevented the "X" close button in the top right of the dialog from showing. This has been resolved, also.

Lastly, there was an issue where the status bar was not displaying content "the best that it could" due to a lot of characters causing some unexpected wrapping - this wrapping behavior has been improved.
image


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/dual-action-dialog-header/dual-action-dialog-header.component.html
```diff
@@ -7,12 +7,12 @@
...
         <span><ng-container *ngIf="!!screenData.headerIcon && !!screenData.headerText">&nbsp;</ng-container>{{screenData.headerText}}</span>
     </div>
-    <app-icon-button 
+    <app-icon-button
         class="back-button" 
         gdArea="backbutton" 
-        *ngIf="screenData.backButton"
-        (buttonClick)="doAction(screenData.backButton)"
-        [actionItem]="screenData.backButton"
+        *ngIf="screenData['doneButton']"
```
jrmihalick, line 13: No changes need in the class file for this?  I don't follow this change.

brandon-lewis: @jrmihalick The done button is being passed in from the state as part of the screen data, so no changes needed to ts components or state classes if that's what you're asking about.

jrmihalick: So it's no longer possible for the screenData to have a `backButton` value?  Could we make a mistake on the server side and set the value of backButton, but then it won't work because now this is expecting doneButton?

brandon-lewis: The issue was the corner X and the sub-menu "back" were both using the same "backButton" property for that navigation, which obviously wasn't working as intended. Both make sense, but I probably should keep the back convention for the top right corner and adjust the component-level nav to use something other than that name. Will make that change now.

jrmihalick (approved):

PR #2096, opened 2022-03-16
On Supported Platforms, Always Try to Auto-Personalize

### LINKED ISSUE #2090: When in a valid ZeroConf environment, only try to auto personalize
When you're on a platform that is supporting ZeroConf it currently attempts to auto personalize once and then falls back to manual personalization techniques. Suggest changing the behavior to only make ZeroConf personalization attempts and only perform a manual personalization attempt after the user explicitly asks it to fallback to a manual mode.

This will solve numerous little data races around like if the server comes up after the client is opened. or the special case that was created just for electron clients running on a specific register.

kdelorey:
Resolves #2090 

Changes the ZeroConf personalization to be modal when in a supported environment. A little dialog will appear that shows the current step that is being tried. This is useful when diagnosing issues over the phone with people, you can easily get feedback on what step its on and translate it back to the stage in code.

Finally if auto-personalization continues to fail for over 90 seconds, a "Manually Connect" button will appear that will skip auto personalization and continue down the manual routes.

image


kumaru-ae:
we can always go to manual personalization by clicking on four corners so that button is redundant and in fact would cause more headache for support

jchirico8 (approved):

PR #2099, opened 2022-03-16
Use persistedId to determine if toast is persistent or not

lutzb-edje:
Use PersistedId to determine if toast is persistent or not. This change is made to assist with the business date mismatch toast


jrmihalick (approved):
Don't forget to update your PR description and link a git hub issue

PR #2109, opened 2022-03-17
wedge character strip from scan or search

### LINKED ISSUE #2108: Wedge scanner input in focused scan-or-search retains control characters
**Describe the bug**
"Bug" where a wedge scanner scanning in a barcode in the `scan-or-search` input still attaches its control characters to the data it sends to the server. This results in oddities like "No results found for *000122521" when scanning a barcode encoded as "000122521".

**To Reproduce**
1. Get a barcode encoded with something, preferably alphanumeric
2. Go to Item Inquiry on a device with a wedge scanner attached that sends some sort of control character (IE: TC52 with DataWedge app)
3. Focus the search bar input (click or tap it)
4. Scan the barcode, and if necessary, press Enter
5. Observe the resulting message stating "No results found for *XXXXXX" where X is your barcode encoded value.

**Expected behavior**
In the short term, these control characters should be stripped from the data like they are in a traditional scan action. In the long term, we may want to consider how to approach wedge scanners in the future. Since they just simulate keypresses really quickly, is this appropriate to be considered a "scanner" in the Commerce sense?

**Screenshots**
Control characters being propagated through to server:
image

Working correctly on IAC:

https://user-images.githubusercontent.com/39058176/158893861-a4489bd9-0fc9-4850-8498-0b294294ebcb.mp4

Working correctly on MPOS/Vetco:

https://user-images.githubusercontent.com/39058176/158894842-4dade5c3-bd2c-411c-bd5e-0bdc73d155b7.mp4

\*Note: I can't prevent the \* here from showing in the search bar. When the keyboard pops, all keypresses become `(unidentified)` and we're unable to tell what key was pressed. Apparently this is a deliberate decision by react. So the \* will still appear, it just never gets sent.

**Additional context**
Relevant Jira:
 - [7793](https://petcoalm.atlassian.net/browse/PDPOS-7793)

JJStarrett:
### Issues Fixed
 - Resolves #2108 

### Summary
In the short term, this strips wedge scanner control characters from `scan-or-search` keyed strings. This is to address a sort of fundamental issue with wedge scanners in that they aren't really scanners in an API sense - more of just automated keyboards. They just simulate keystrokes very quickly. So we use control characters and a short time out to detect when they're "scanning". However, obviously, it's just keystrokes and there's nothing really to stop them from "scanning" into input fields. I believe recently there was an issue with customer loyalty due to this.

This is compounded by the fact that where this wedge scanner is used - Android mobile TC devices - the keyboard deliberately obscures its keypresses while it's popped. So when the scanner simulates those presses, we just get a whole bunch of `(unidentified)` keys detected. So we can't "see" a control character as it's pressed. We can only see the string afterwards if an input is focused (keyboard is popped). If you look at the linked issue, that's why the \* character is still in the search box on the second video.

"Scan" with react keyboard popped: 
image

"Scan" with react keyboard not popped:
image

In the long term, we may want to consider how to approach wedge scanners in the future. Since they just simulate keypresses really quickly, is this appropriate to be considered a "scanner" in the Commerce sense? Or maybe we should have keystrokes autofocus the most relevant input?

---

This also adds some additional logging in `scan-part` for https://petcoalm.atlassian.net/browse/PDPOS-7785


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/barcode-scanners/wedge-scanner/wedge-scanner.plugin.ts
```diff
@@ -194,13 +194,31 @@ export class WedgeScannerPlugin implements Scanner {
                 if (key === 'Enter') {
                     key = '\n';
                 }
-                charList.push(e.key);
```
jrmihalick, line 197: There is another parallel PR with this change, you can remove this.

jrmihalick (changes requested):
Should the existing change for the Loyalty component be adapted to use your changes here?

jrmihalick:
Also what is the behavior on screens where we ran the default action because the scan is terminated by an Enter key?

JJStarrett:
> Also what is the behavior on screens where we ran the default action because the scan is terminated by an Enter key?

This should only affect screens with scan-or-search elements (item inquiry, manage users, order pick, transaction search, central price checker, and sometimes sale or return), but do you just mean if we terminate with 'Enter' instead of 'Tab', what happens? I've got two videos in the ticket that should show the two scenarios on item inquiry specifically, but both should be unaffected. We're only filtering the string data after the keystrokes are entered here.

Is there a specific screen you wanted me to validate?

JJStarrett:
As per standup, going to try new solution of adding scan-part to loyalty customer dialog and see if that works here too. This will remove the auto-focus on the customer dialog prompt and potentially drop support for wedge scanning directly into an input field

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/barcode-scanners/wedge-scanner/wedge-scanner.plugin.ts
JJStarrett: Gotcha, Removed

JJStarrett:
> Should the existing change for the Loyalty component be adapted to use your changes here?

Makes sense, has been adapted

jrmihalick (approved):

PR #2113, opened 2022-03-21
Add new dialog for EWallet tender flow

jchirico8:
### Issues Fixed
JumpMind/commerce#5813

image
image
image


chenson42:
@kdelorey @jchirico8 Was there anything we needed to add the the e wallet dialog to support scandit?

kdelorey:
> @kdelorey @jchirico8 Was there anything we needed to add the the e wallet dialog to support scandit?

Scandit has QR codes turned on, so the data will be sent to the server, but its up to the server to handle that data appropriately.

Does Scandit even matter here? It should be the customer scanning this QR code with their own device/app right?

kdelorey:
> It's supposed to do both. We show a QR code for the user to scan, but it doesn't actually work right now so you have to scan the users barcode

Ah, okay. Do you know what kinds of barcode symbologies to expect, I can make sure scandit will be ready for it?

jchirico8:
When I went to my Venmo it had a regular barcode and a qr code. I'm not sure how picky their sandbox environment is though with the value you send

chenson42:
@kdelorey I just meant to ask if there was a component that we needed to include in the dialog in order to get the scandit button and subsequent window.

kdelorey:
> @kdelorey I just meant to ask if there was a component that we needed to include in the dialog in order to get the scandit button and subsequent window.

Ah yeah, there's a `image-scanner.component` that you can use stand alone. There are no good examples of this being used stand alone, it's typically paired with an input prompt or the `scan-or-search.component`. But it should be as simple as popping the component down on a surface with a fixed size and hooking the `scan` event up to an action handler.

You need to handle the case without a image scanner (aka fixed register). In which case you use the `scan-part.component` to activate whatever the optical scanner is.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/e-wallet-payment-dialog/e-wallet-payment-dialog.component.html
```diff
@@ -0,0 +1,17 @@
...
+        <mat-dialog-actions *ngIf="screen.cancelButton" class="dialog-actions">
+            <app-secondary-button (click)="doAction(screen.cancelButton)">
+                <span responsive-class>{{screen.cancelButton.title}}</span>
+                <app-icon *ngIf="screen.cancelButton.icon" [iconClass]="'md'" [iconName]="screen.cancelButton.icon">
+                </app-icon>
+            </app-secondary-button>
+        </mat-dialog-actions>
+    </div>
+</div>
+<app-scan-part></app-scan-part>
```
kdelorey, line 17: it looks like you have the optical scanner active, but no support for image scanning which is the `app-image-scanner`.

It needs a fixed-ish sized area to work from and you should only image scan if its supported (see `barcode-scanner.service.ts`)

kdelorey:
@jchirico8 did you need me to try it on the iPad?

kdelorey (approved):

PR #2122, opened 2022-03-24
Re-Init scanner settings when config change message received from bac…

shiljo-litmus7:
Issues Fixed
[PDPOS-7878](https://petcoalm.atlassian.net/browse/PDPOS-7878)

[3.0][Mobile Grooming][mPOS Vetco] - Intermittent issue with scanner (Tracking)

Summary
Scanning is not working properly in MPOS Vetco devices which uses wedge scanner(Used by ScanOrSearchComponent). This happens as the scan settings are done during component construction and doesnt get updated even if a different Connfig message is received from back-end. Because of this non-default scan config never gets applied


jrmihalick (commented):
I need to better understand what you are seeing and we need to identify what precisely changed between 1.13 and master to cause this since this was working fine in 1.13.

jrmihalick (approved):

PR #2125, opened 2022-03-24
Fix 'horizontal scroll on combobox form field #2124'

### LINKED ISSUE #2124: horizontal scroll on combobox form field
**Describe the bug**
The dynamic form has an unnecessary horizontal scroll on the combo box fields (mat-select)

**Screenshots**
horizontal-scroll-on-form-field

stojankz:
This issue is a result of a style that is applied directly to the mat-select by an angular animation component that set min-width as calc(100% + 32px). Because the style was on the element directly and not a class, I was not able to find a cleaner way to override it than with the use of !important.

Screen Shot 2022-03-24 at 12 25 10 PM

Link to angular animation that resulted in this bug: https://github.com/angular/components/blob/59002e1649123922df3532f4be78c485a73c5bc1/src/material/select/select-animations.ts

### Issues Fixed
Closes #2124

### Screenshot
horizontal-scroll-on-form-field-fix


jchirico8 (approved):

jrmihalick (approved):

jrmihalick:
Nice hack

https://media.giphy.com/media/IkGlH9PsraE4YTKDHD/giphy.gif

PR #2135, opened 2022-03-28
Loyalty Signup Button not Updating from Customer Display

blubaugha:
### Issues Fixed
**Loyalty Signup Button not Updating from Customer Display**
[2134](https://github.com/JumpMind/openpos-framework/issues/2134)


openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/LoyaltySignupMessage.java
```diff
@@ -12,6 +13,7 @@
 public class LoyaltySignupMessage extends Message {
     private static final long serialVersionUID = 1L;
 
+    @JsonProperty("isActiveOnCustomerDisplay")
```
kdelorey, line 16: I'm guessing before it would export as `activeOnCustomerDisplay`? I wonder if it would be bettered served fixing on the client side? Not sure, I think JS names tend to avoid leading `is` and `has`.

kdelorey (approved):

Issue #2136, opened 2022-03-28
Membership - Non-Member Shows Blank Name


blubaugha:
**Description**
When a customer is linked, a list of memberships show on the sale screen and customer details screen. For all groups the customer does not belong to, a blank name is displayed in the UI pill.

PR #2137, opened 2022-03-28
Membership - Non-Member Shows Blank Name

blubaugha:
### Issues Fixed
**Membership - Non-Member Shows Blank Name**
[2136](https://github.com/JumpMind/openpos-framework/issues/2136)


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/membership-display/membership-display.component.html
```diff
@@ -12,7 +12,7 @@
         <div *ngIf="!membership?.member && membership?.nonMemberName" class="membership-name">
             {{membership.nonMemberName}}
         </div>
-        <div *ngIf="membership?.member" class="membership-name">
+        <div *ngIf="membership?.member || !membership?.nonMemberName" class="membership-name">
```
kdelorey, line 15: `*ngIf` supports a `else` which would probably be useful here maybe.

```html
<div *ngIf="!membership?.member && membership?nonMemberName; else membershipName">
    Some Content Here
</div>
<ng-template #membershipName>Some Other Content</ng-template>
```

kdelorey (approved):

PR #2153, opened 2022-04-01
2.7 to international

r-cochran:
pulling 2.7 changes into branch


r-cochran:
@jrmihalick @kdelorey one of the trickier parts of this merge was both of you worked on auto personalization, but in different ways. Would appreciate both of you taking a look at this.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/zeroconf/zeroconf.ts
```diff
@@ -6,6 +6,8 @@ export const ZEROCONF_TOKEN = new InjectionToken<Array<Zeroconf>>('ZEROCONF');
 export interface Zeroconf {
     watch(type: string, domain: string): Observable<ZeroconfResult>;
 
+    deviceName(): Observable<string>;
```
kdelorey, line 9: I think we were supposed to remove this method in favor of it being apart of like platform service or something like that.

jchirico8 (commented):
It looks like `keypress.provider.ts` was deleted in master/international but it's showing here as added back but empty.

jchirico8 (dismissed):

kdelorey (approved):

Issue #2156, opened 2022-04-04
Auto personalization ignores SSL setting in dev_device_personalization.ssl_enabled_flag


jrmihalick:
**Describe the bug**
The value of `dev_device_personalization.ssl_enabled_flag` has no affect on whether or not SSL is used upon personalization.  It just always uses http.

**To Reproduce**
1. Set up a row in dev_device_personalization with ssl_enabled_flag = 1
2. Try to auto personalize

Notice that the device will attempt to personalize over http instead of https.

**Expected behavior**
The device should auto personalize with https when dev_device_personalization.ssl_enabled_flag=1

**Additional context**
- PDPOS-7974: [3.0] Mobile app Auto-personalization is not working

PR #2157, opened 2022-04-04
Use the value of dev_device_personalization.ssl_enabled_flag when auto personalizing

jrmihalick:
I already fixed this once in 1.13

### Issues Fixed
- #2156 
- PDPOS-7974: [3.0] Mobile app Auto-personalization is not working


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/personalization/personalization.service.ts
```diff
@@ -150,7 +150,6 @@ export class PersonalizationService {
                 tap(response => {
                     if (response) {
                         console.info(`Auto personalization response received: ${JSON.stringify(response)}`);
-                        response.sslEnabled = this.sslEnabled$.getValue();
```
jrmihalick, line 153: `sslEnabled` needs to come from the DB and therefore the auto personalization response.

chenson42 (approved):

jchirico8 (approved):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/personalization/personalization.service.ts
kdelorey: I added a way for for at least the zeroconf service discovery method to specify during the discovery that it should call the `personalizeMe` endpoint over `https`. 

Anyways, I just wanted to mention that. Nothing else to report here.

jrmihalick: @kdelorey So this change will break that for you?

kdelorey: No its not even used, I just left the door open. This assignment happens after `personalizeMe` is called anyways.

PR #2158, opened 2022-04-05
Reset paired device when shrug happens

jchirico8:
I added a reset function to the container because `stateManagersByDeviceId` wasn't publicly accessible. Alternative options would be to add a getter for the whole map or a `getStateManagerByDeviceId`. If those are preferred I can update this PR.

### Issues Fixed
Jumpmind/commerce#5956

customer-display-reset


stojankz (approved):

chenson42 (approved):

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/ErrorGlobalActionHandler.java
```diff
@@ -9,8 +11,14 @@
...
     IStateManager stateManager;
 
+    @Autowired
+    IStateManagerContainer stateManagerContainer;
+
     @OnGlobalAction
     public void onResetStateManager(Action action) {
         stateManager.reset();
+        if (StringUtils.isNotBlank(stateManager.getPairedDeviceId())) {
+            stateManagerContainer.resetStateManager(stateManager.getPairedDeviceId());
```
jrmihalick, line 21: @JJStarrett How does this look to you for the e355?  This will reset the e355 statemanager right?

JJStarrett: Right, we already effectively reset it on client disconnection, though. I don't think this would cause any issues with the E355.

JJStarrett (approved):

PR #2159, opened 2022-04-05
Previously delivered config message from server for Wedge scanner is retrieved and applied

shiljo-litmus7:
Issues Fixed
[PDPOS-7878](https://petcoalm.atlassian.net/browse/PDPOS-7878)

[3.0][Mobile Grooming][mPOS Vetco] - Intermittent issue with scanner (Tracking)

Summary
Wedge scanner config messages for certain mobile devices(Vetco, Mobile Grooming) are different from what is the default in front-end codebase. The relevant config  is sent from back-end which sometimes get lost as the config subscription sometimes happens after message delivery.

The change is to use the ReplaySubject which will ensure that previous messages will get delivered to the subscriber. There is already a ConfigurationService which provides this Replay capability and it is leveraged


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/barcode-scanners/wedge-scanner/wedge-scanner.plugin.ts
```diff
@@ -52,8 +55,10 @@ export class WedgeScannerPlugin implements Scanner {
                 this.timeout = m.timeout;
             }
 
+            console.debug('Received config message for WedgeScanner.Re-Creating scan observable');
             // Re-Initialize the scan with updated config received from server
             this.scanObservable = this.createScanObservable();
+            console.debug('Received config message for WedgeScanner.Re-Created scan observable');
         });
 
         sessionService.getMessages('ConfigChanged').pipe(
```
jrmihalick, line 64: Go ahead and update this to use the config service also for WedgeScannerTypes

jrmihalick (changes requested):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/barcode-scanners/wedge-scanner/wedge-scanner.plugin.ts
shiljo-litmus7: Done

jrmihalick (approved):

Issue #2162, opened 2022-04-06
Wedge scanner keyboard event listeners causing client slowdown


JJStarrett:
**Describe the bug**
When leaving the client idle for long periods of time, it tends to slow down while showing increasing amounts of keyboard events in the client log from the wedge scanner plugin.

**To Reproduce**
1. Enter the sale screen so the loyalty prompt displays
2. Open the chrome devtools and look for the `Start/Stop key ( ) pressed:` log message when typing some character into the prompt. Observe how many times it displays.
3. Switch to another tab and leave the device idle for ~3 minutes
4. Switch back to Commerce tab and ensure the "Connecting to Server..." dialog briefly comes up
5. Ensure you're still on the loyalty prompt and repeat step 2.
6. The log statement occurrences should be greater than last time.

**Expected behavior**
We should not gradually be increasing the amount of keyboard event listeners every time we reconnect to the server

**Screenshots**
After 2 reconnects:
<img width="1154" alt="two_reconnects" src="https://user-images.githubusercontent.com/39058176/162025105-08b7ee9c-f221-43e5-9a67-7045c3bf9468.PNG">

After 3 reconnects:
<img width="1160" alt="three_reconnects" src="https://user-images.githubusercontent.com/39058176/162025140-b45780df-71d3-4917-a639-5bc5cea958ba.PNG">

After 4 reconnects:
<img width="1179" alt="four_reconnects" src="https://user-images.githubusercontent.com/39058176/162025182-1039a7f1-7116-43ff-83f8-6634eda054b8.PNG">

**Additional context**
Relevant Jira ticket:
 - [7828](https://petcoalm.atlassian.net/browse/PDPOS-7828)

PR #2163, opened 2022-04-06
Don't recreate new keyboard event listeners

JJStarrett:
### Issues Fixed
 - #2162 

### Summary
We appear to be gradually, and sometimes suddenly, accumulating multiple keyboard event listeners as the client reconnects to the server after remaining idle for a while. For example, this is what happens after leaving the client idle such that it eventually reconnects to the server:

After 2 reconnects:
two_reconnects

After 3 reconnects:
three_reconnects

After 4 reconnects:
four_reconnects

I suspect this is because we are continually recreating keyboard event listeners and then subscribing to them without ever unsubscribing from the old observables. I'm not positive though. However, this does appear to fix the problem, as we only have a consistent 2 event notifications in the client log with these changes:

After 2 reconnects:
two_reconnects_alt

After 3 reconnects:
three_reconnects_alt

After 4 reconnects:
four_reconnects_alt


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/barcode-scanners/wedge-scanner/wedge-scanner.plugin.ts
```diff
@@ -51,6 +58,9 @@ export class WedgeScannerPlugin implements Scanner {
             if (m.timeout) {
                 this.timeout = m.timeout;
             }
+            if (m.enabled) {
```
jrmihalick, line 61: `this.enabled` defaults to `true`, but if `m.enabled = false`, then this.enabled will never be set to false?  Or am I missing something?

jrmihalick (dismissed):

JJStarrett:
@kdelorey Could we get your thoughts on these changes?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/barcode-scanners/wedge-scanner/wedge-scanner.plugin.ts
```diff
@@ -71,69 +79,89 @@ export class WedgeScannerPlugin implements Scanner {
...
-            const subscription = this.createScanBuffer().subscribe({
-                next: d => observer.next(d),
-            });
-
-            return () => {
-                subscription.unsubscribe();
-                this.scannerActive = false;
-            };
-        }).pipe(
+        if (this.enabled) {
```
kdelorey, line 82: Its probably fine here, but I figured I'd mention it so you're thinking about it, its usually a good idea to try to stick to using an observable chain for the whole method. The reason for this is that they are often COLD, meaning none of these observables are actually ever executed until something subscribes for the first time. But since `this.enabled` is checked outside of the observable sequence, it could have changed between when the observable was created and when it was actually subscribed. This is very similar to your standard lazy initialization.

So for this example it might be good to look at the `enabled` flag only when someone subscribes. Which would look something like this:

```ts
return iif(
    () => this.enabled,
    
    // true
    new Observable(observer => {
        // do true stuff
    }),
    
    // false
    new Observable(observer => {
       // do false stuff
    })
);
```

kdelorey (dismissed):
lgtm

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/platform-plugins/barcode-scanners/wedge-scanner/wedge-scanner.plugin.ts
JJStarrett: Oh awesome, I will implement this! I appreciate the feedback on this, Kevin. The analogy to lazy initialization makes sense to me.

jrmihalick (approved):

jrmihalick (approved):

PR #2167, opened 2022-04-08
Create endpoint to change a device's brand

jchirico8:
### Issues Fixed
JumpMind/commerce#6147


chenson42:
In hindsight, brand should probably not be a tag.   I wonder if we should pull off the bandaid and move it to a column on dev_device (versus the configurable tag)?

@mmichalek @gwilmer @jchirico8

jchirico8:
Right now it's duplicated as a tag and device param. I can remove the tag and keep the param so we don't add a column that's not needed for other customers @chenson42

mmichalek:
I think every customer does have at least one brand, so think we should make it a core concept in the system.

Also - what is the role of device params?  They seem confusing as to what they are for vs. tags.

jrmihalick:
> I think every customer does have at least one brand, so think we should make it a core concept in the system.
> 
> Also - what is the role of device params? They seem confusing as to what they are for vs. tags.

Device params are used for storing the values given on the 3rd step of personalization UI.  E.g., `brandId`, `deviceType`, `businessUnitId` (for vetco).  Those params eventually end up on the `StateManager.personalizationProperties` and in the `ClientContext`.  They also get passed over to the paired device StateManager.

There is some overlap on tags, but I guess that is by coincidence.  The only overlap I see is the deviceType and perhaps the brand.  But I think the brandId given on personalization is what is really used for choosing resources.

mmichalek:
> `brandId`, `deviceType`, `businessUnitId`

It seems like if businessUnitId and brandId are built in values on dev_device, and deviceType is a tag ... that these concepts could be merged.

jrmihalick:
> > `brandId`, `deviceType`, `businessUnitId`
> 
> It seems like if businessUnitId and brandId are built in values on dev_device, and deviceType is a tag ... that these concepts of tag and device param could be merged.

I think the original idea was to allow for custom parameters on personalization as well.  The params are configurable in the `openpos.ui.personalization.parameters` config property.  I added `businessUnitId` as a "recognized" personalization parameter when working on vetco because I wasn't sure if the store id component of the device id would necessarily always be the same as the businessUnitId, but I guess we have settled on them being the same thing.

jchirico8:
We're going to hold off on changes to the DeviceModel in 2.x

PR #2191, opened 2022-04-15
Port code to look up manifest version from symds

chenson42:
Port code to look up manifest version from symds.   We has inherited a change we didn't want from 3.13.   Moving forward with our own version.


stojankz (approved):

kdelorey (approved):
Oh interesting. This could be good for the updater version interface too.

jchirico8 (approved):

PR #2196, opened 2022-04-18
hard coded strings update

r-cochran:
### Issues Fixed
[5823](https://github.com/JumpMind/commerce/issues/5823)


openpos-flow/src/main/java/org/jumpmind/pos/core/flow/ui/PromptConfig.java
```diff
@@ -24,7 +24,7 @@ public class PromptConfig {
     private BigDecimal min;
     private BigDecimal max;
     private DialogProperties dialogProperties;
-    private ActionItem actionMenuItem = new ActionItem("Next", "Next", true);
+    private ActionItem actionMenuItem = new ActionItem("key:common:action.next.label", "Next", true);
```
jchirico8, line 27: Isn't the first param the action and the second the title?

jchirico8: jk on 3 of the constructors someone switched the order of action and title 🤦

jchirico8 (approved):

openpos-flow/src/main/java/org/jumpmind/pos/core/flow/ui/PromptConfig.java
r-cochran: yeaaaaaaaaaaah. there are like 8 constructors for that class and checking each one needed constant reminding of which one was the actual title.

jchirico8 (approved):

PR #2208, opened 2022-04-26
Bugfix/pdpos 8142

anandlalvb:
### Issues Fixed
Utility function to covert timezone to offset and vice versa.
Related to issue: https://github.com/JumpMind/commerce/issues/6342
PR: https://github.com/JumpMind/commerce/pull/6343


openpos-util/src/main/java/org/jumpmind/pos/util/AppUtils.java
```diff
@@ -151,4 +153,28 @@ public static Date getTimezoneOffsetCorrectedDate(Date serverDate, String server
         }
         return serverDate;
     }
+
+    /**
+     * Converts offset to millis
+     *
+     */
+    public static Integer convertOffsetToMillis(String offsetString) {
```
jrmihalick, line 161: Might as well just return an `int` instead of Integer?

openpos-util/src/main/java/org/jumpmind/pos/util/AppUtils.java
```diff
@@ -151,4 +153,28 @@ public static Date getTimezoneOffsetCorrectedDate(Date serverDate, String server
...
+        return (int) (clientOffset.get(ChronoField.OFFSET_SECONDS) * 1000L);
+    }
+
+    /**
+     * Converts offset in millis to timezone
+     *
+     * @param offsetMillis
+     * @return
+     */
+    public static String getTimezone(Integer offsetMillis){
```
jrmihalick, line 172: Might as well just accept an `int` instead of an `Integer` since it gets unboxed anyway in the first line of code?

openpos-util/src/main/java/org/jumpmind/pos/util/AppUtils.java
```diff
@@ -151,4 +153,28 @@ public static Date getTimezoneOffsetCorrectedDate(Date serverDate, String server
...
+        int absOffset = Math.abs(offsetMillis);
+        String strOffset = String.format("%s%02d:%02d", offsetMillis < 0 ? "-" : "+",
```
jrmihalick, line 174: This line that creates the string offset is also shared with `getTimezoneOffsetCorrectedDate(Date serverDate, int serverOffsetMillis, String clientOffsetString)`.  Could you pull out the logic into another private method and share it instead of duplicating it?

jrmihalick (changes requested):

jrmihalick (approved):

PR #2211, opened 2022-05-02
Create annotation to skip showScreen logging of UIMessage

jchirico8:
### Issues Fixed
JumpMind/commerce#6392


openpos-flow/src/main/java/org/jumpmind/pos/core/ui/message/CustomerDisplayHomeNoLogUIMessage.java
```diff
@@ -0,0 +1,7 @@
+package org.jumpmind.pos.core.ui.message;
+
+import org.jumpmind.pos.util.SuppressScreenLogging;
+
+@SuppressScreenLogging
```
jchirico8, line 5: I don't know if I'm sold on this being the best way. Another option would be to just set a flag on UIMessage itself to skip this logging.

chenson42 (approved):

PR #2293, opened 2022-06-13
Fix 'Screens shown out of order when same screen is sent to the client too close together. #2292'

### LINKED ISSUE #2292: Screens shown out of order when same screen is sent to the client too close together.
**Describe the bug**
When two screens are sent too close together, there is a case where the first screen stops to wait for a dialog to close before displaying. The second screen then finishes “displaying” while the first screen is waiting and the first screen gets shown last after the dialog has been confirmed to have been closed.

ie in scenario below: when an item is searched, there is a loading dialog that is shown before the first screen (where the availability is still PENDING). The availability is determined and the screen is re-sent quickly after the first screen. Because the second screen does not have to wait for dialog closing, it ends up showing first followed by the initial screen which is incorrect.

**To Reproduce**
Steps to reproduce the behavior while running CommerceDemo- Fashion :
1. Go to Item Inquiry
2. Search '0105-1165'
3. See issue.

**Screenshots**
order-button-disabled-bug-commerce

stojankz:
### Issues Fixed
Closes #2292 - When two screens are sent too close together, there is a case where the first screen stops to wait for a dialog to close before displaying. The second screen then finishes “displaying” while the first screen is waiting and the first screen gets shown last after the dialog has been confirmed to have been closed.

### Notable Mention
Shoutout to @jchirico8 who helped me debug and find the root of the problem. 

### Screenshots
Original Bug:
order-button-disabled-bug-commerce

Fix:
order-button-disabled-bug-commerce-fix


jchirico8 (dismissed):

jrmihalick (approved):
Only potential issue I see here is if a client component shows a screen without going through the server side, but in looking around the client code I don't see any cases of that in the code any longer.  Only thing like it is when we show the Loading screen from the client side, but that has special handling in the dynamic-screen component and isn't handled like other messages.

jchirico8 (approved):

PR #2301, opened 2022-06-14
Fix burl/0.9 deploy stage

MRUFFING67:
Fixing the deploy stage for burl-0.9. Somethign is wrong with executing the upload archives task.

-Note: Chris told me not to tie this to an issue number.


MRUFFING67:
The newer versions of Openpos use the parameter "-PdeploySftpUrl" but this older version needs "-PdeployUploadUrl".

Fixed that in the Jenkinsfile deploy stage, but can only be executed once it is merged with this releases branch...

chenson42 (approved):

PR #2335, opened 2022-06-29
Be consistent where we track training mode.   No need to put it on the device model when it is in ClientContext

chenson42:
Be consistent where we track training mode.   No need to put it on the device model when it is in ClientContext


chenson42:
https://github.com/JumpMind/commerce/issues/6899

stojankz (approved):

Issue #2345, opened 2022-07-05
AECC lookup - Print Temp Pass and close button displayed in left side


sivasubrad:
<img width="367" alt="Screen Shot 2022-07-05 at 1 10 32 PM" src="https://user-images.githubusercontent.com/103125078/177380759-c8b056c8-360c-4516-94d7-a0a005556053.png">

Print Temp Pass and close button are displayed on the left side of the dialog instead of right only in mobile device.

[https://jira.ae.com/browse/PCOM-9062](url)

PR #2346, opened 2022-07-05
PCOM-9062 Print Temp Pass and close button align on right

sivasubrad:
Jira : [https://jira.ae.com/browse/PCOM-9062](url)

Issue : [https://github.com/JumpMind/openpos-framework/issues/2345](url)

Print Temp Pass and close button are displayed on the left side of the dialog instead of right only in mobile device.

Couldn't replicate this in SIT but happening only in mobile devices in lab. So trying to add right alignment style with a class in html.

Screen Shot 2022-07-05 at 1 15 59 PM

### DIFF
--- a/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/customer-details-dialog/customer-details-dialog.component.scss
+++ b/openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/customer-details-dialog/customer-details-dialog.component.scss
@@ -29,7 +29,7 @@
   }
 }
 
- .account-detail-buttons{
+ .customer-detail-buttons{
     justify-content: end;
     align-content: end;
   }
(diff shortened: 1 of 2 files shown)


openpos-client-libs/projects/openpos-client-core-lib/src/lib/screens-with-parts/customer-details-dialog/customer-details-dialog.component.scss
```diff
@@ -29,6 +29,10 @@
   }
 }
 
+ .account-detail-buttons{
+    justify-content: end;
```
jchirico8, line 33: the buttons class has these same styles 4-5 lines below this. why do we need to redefine the same style to get the buttons to move over?

sivasubrad: Mobile devices still have them on the left side. I am not able to reproduce it in SIT but it is happening in lab. Since button class is used in multiple places, just trying to have a unique class for the right alignment.

jchirico8 (changes requested):

PR #2366, opened 2022-07-12
added converter for BigDecimal to Money for ReflectUtils

### LINKED ISSUE #2365: ReflectUtils add BigDecimal to Money conversion
Need the ability for converting BigDecimal to Money as part of the ReflectUtils process.

**Additional context**
PCOM-8710

r-cochran:
### Issues Fixed
#2365


openpos-util/src/main/java/org/jumpmind/pos/util/ReflectUtils.java
```diff
@@ -69,6 +72,10 @@ private static Object coerceValue(Field field, Object value) {
             value = ((Number) value).intValue();
         } else if (field.getType().equals(BigDecimal.class) && value instanceof Number) {
             value = new BigDecimal(value.toString());
+        } else if (field.getType().equals(Money.class) && value instanceof BigDecimal) {
```
kdelorey, line 75: As someone dealing with a ton of foreign currency issues at the moment, I don't really like this as the currency is assumed USD.

Side note since you lack context of usage, maybe prefer half even rounding.

mmichalek: Based on Kevin's comment, I'm not sure that ReflectUtils can be smart enough to go from BigDecimal to Money, which requires "business context" knowledge of the currency

r-cochran: Yeah i ran into the missing context as to what the actual currency is supposed to be in this scenario. Ended up having to add behavior after the query has completed to basically set the right currency on the money objects. Not ideal for sure. Would it make sense to pass some kind of context object down into the query logic that can be referenced at this point?

chenson42 (approved):

PR #2377, opened 2022-07-18
Fixed issue where personalizing an existing device would reset TAG_ values

kdelorey:
Resolves JumpMind/commerce#6910


openpos-devices/src/main/java/org/jumpmind/pos/devices/service/PersonalizeEndpoint.java
```diff
@@ -49,19 +49,19 @@ public PersonalizationResponse personalize(@RequestBody PersonalizationRequest r
         final String parentAppId = request.getParentAppId();
         final String parentDeviceId = request.getParentDeviceId();
 
-        DeviceModel deviceModel;
+        DeviceModel deviceModel = null;
 
         if (isNotBlank(deviceId) && isNotBlank(appId)) {
             try {
                 log.info("Validating auth request of {} as {}", deviceId, appId);
+                deviceModel = devicesRepository.getDevice(deviceId);
```
jrmihalick, line 57: This could throw a DeviceNotFoundException here whereas it wouldn't have done so before, which sounds like correct behavior.  However, for a virtual device, are we still ok?  I am not sure.  @chenson42 should probably look since he worked on that.

kdelorey: Device not found would be seen as a new device and the exception is caught and responds accordingly. As far as virtual devices, they have a sort of special repository underneath and the main repo effectively joins the results of the DB and the special repo together.

chenson42: i think it looks good

PR #2395, opened 2022-07-22
Add PR related methods

MaximAndreevich:
New methods created to store more device related parameters
We will use it to keep Puerto Rico device ID that we use to generate a fiscal number.


openpos-devices/src/main/java/org/jumpmind/pos/devices/model/DevicesRepository.java
```diff
@@ -261,6 +261,27 @@ private List<DeviceParamModel> getDeviceParams(String deviceId) {
...
+        params.put("deviceId", deviceId);
+        params.put("paramName", paramName);
+        DeviceParamModel response = devSession.findByFields(DeviceParamModel.class, params, 10000).stream()
+                .findFirst()
+                .orElse(null);
+        return Objects.nonNull(response) ? response.getParamValue() : null;
+    }
+
+    public void saveNewParam(DeviceParamModel model) {
+        devSession.save(model);
```
jchirico8, line 282: Could you call this from `saveDevice`?

openpos-devices/src/main/java/org/jumpmind/pos/devices/model/DevicesRepository.java
```diff
@@ -261,6 +261,27 @@ private List<DeviceParamModel> getDeviceParams(String deviceId) {
         return devSession.findByFields(DeviceParamModel.class, params, 10000);
     }
 
+    public List<DeviceParamModel> getAllDeviceParamsByParamName(String paramName) {
```
jchirico8, line 264: This class is well covered by unit tests. Can you write new tests to cover these new functions?

jchirico8 (commented):
You might need to create new endpoints to hit these new functions. I'm not familiar with how you're going to access these right now or if we have a concrete standard on whether other services should call IDevicesService to get anything from the DevicesRepository or if you can wire the repository directly into another endpoint to use

openpos-devices/src/main/java/org/jumpmind/pos/devices/model/DevicesRepository.java
MaximAndreevich: I changed logic and call saveDevice now

openpos-devices/src/main/java/org/jumpmind/pos/devices/model/DevicesRepository.java
MaximAndreevich: Created

openpos-devices/src/test/java/org/jumpmind/pos/devices/model/DevicesRepositoryTest.java
```diff
@@ -328,6 +324,24 @@ public void testUpdateDeviceStatusDifferentDeviceId() {
...
 
+    @Test
+    public void testGetAllParamsByParamName(){
+        Map<String, Object> paramNamesMap = new HashMap<>();
+        paramNamesMap.put("paramName", "TestParam");
+        DeviceParamModel param = DeviceParamModel.builder()
+                .paramName("TestParam")
+                .deviceId(mockDeviceId)
+                .paramValue("TestValue")
+                .appId("pos")
```
jchirico8, line 335:
```suggestion
                .appId(mockAppId)
```

jchirico8 (approved):

PR #2399, opened 2022-07-22
Added manual formatter control for some fields

### LINKED ISSUE #2398: Provide a more manual way to control the formatter
The formatters are typically implicitly controlled by prompt types but in some instances we need to use the formatters outside of prompts or there is a need to mix the prompting. Additionally there is a need to provide additional parameters to the formatters such as controlling the currency of the money formatter.

kdelorey:
Resolves #2398

* Money formatter can take argument to control currency
* Money formatter can also recognize formats like `USD 0.00` to format as `$0.00`
* Money formatter can detect if the currency is foreign and if so it uses the long format with the country code.


openpos-flow/src/main/java/org/jumpmind/pos/core/ui/data/SellItem.java
```diff
@@ -1,420 +1,425 @@
...
+                subtitle = String.format(subtitle, this.getPosItemId(), altItemId, this.getQuantity(), this.getSellingPrice(),
+                        this.getSalesAssociate());
+            } else {
+                subtitle = String.format(subtitle, this.getPosItemId(), altItemId, this.getQuantity(), this.getSellingPrice());
+            }
+        }
+        return subtitle;
+    }
+
+    public String getOrderIcon() {
```
jchirico8, line 66: Did you want to clean up these getters and setters since you already did that to PromptUIMessage?

kdelorey: Oh yeah, that makes sense I can do that.

kdelorey:
I've rebased the PR and cleaned up SellItem.

jchirico8 (approved):

PR #2417, opened 2022-07-29
So long and thanks for all the fish

kdelorey:
Removing myself as a code owner


jchirico8 (approved):
Oh I can get on board with this change. Easy approval!

PR #2432, opened 2022-08-09
Feature/upgrade TSLint to ESLint for Openpos

MRUFFING67:
Upgrades (the now-deprecated) TSLint to ESLint for Openpos-framework.
Adapts most rules from TSLint to the version in ESLint, besides those outlined in the attached .txt file, which also details some change in functionality:

[tslint-to-eslint-config.txt](https://github.com/JumpMind/openpos-framework/files/9294053/tslint-to-eslint-config.txt)

(Yes, 350 changed files is correct)


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/formatters/non-numeric.formatter.spec.ts
```diff
@@ -14,7 +16,7 @@ describe('NonNumericFormatter Test', () => {
     });
 
     it('#ensure special characters are not removed', () => {
-        // tslint:disable-next-line:quotemark
+        // eslint-disable-next-line quotes, @typescript-eslint/quotes
```
jchirico8, line 19: Delete this and fix the quotes. I can't believe someone thought it was easier to add a disable line than to just replace the double quotes

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/sale-loyalty-part/sale-loyalty-part.component.spec.ts
```diff
@@ -79,12 +79,12 @@ describe('SaleLoyaltyPartComponent', () => {
                     expect(component.keybindsEnabled(menuItem)).toBe(false);
                 });
 
-                it(`returns false when CONFIGURATION.enableKeybinds is true and the menuItem.keybind is truthy and is 'Enter'`, () => {
+                it('returns false when CONFIGURATION.enableKeybinds is true and the menuItem.keybind is truthy and is \'Enter\'', () => {
```
jchirico8, line 82: oof I'm not a fan of this. are backticks banned altogether or they're just only allowed when you interpolate a variable in the middle of a string?

openpos-client-libs/projects/openpos-client-core-lib/src/lib/utilites/deep-assign.spec.ts
```diff
@@ -25,8 +25,10 @@ describe('deepAssign', () => {
 
     it('Should return the src object if target is undefined', () => {
         const src = { prop1: 'blah', prop2: 4 };
+        /* eslint-disable prefer-const */
         let target;
```
jchirico8, line 29: This should be one line `const target = deepAssign` instead of 2 with a disabled rule

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/sale-loyalty-part/sale-loyalty-part.component.spec.ts
MRUFFING67: @jchirico8 Are you saying you want these to be backticks? The rule comes from @typescript-eslint/quotes rule and any/all string literals in a typescript file will be marked as an error if not surrounded by single-quotes.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/utilites/deep-assign.spec.ts
MRUFFING67: I tried exactly that and IntelliJ does not like it. 'target' is being passed into deepAssign before/during its declaration. The point of the test is to have 'target' be undefined before being passed to deepAssign, which will then reassign target to 'src', so target shouldn't be const in the context of this test.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/sale-loyalty-part/sale-loyalty-part.component.spec.ts
jchirico8: yeah I don't like having to escape quotes in the string when you can just use a different character to make the string

openpos-client-libs/projects/openpos-client-core-lib/src/lib/utilites/deep-assign.spec.ts
jchirico8: hmm the variable is doing 2 things then. could you do `deepAssign(undefined, src)` since that's what it sounds like the test is trying to do?

jchirico8 (approved):

PR #2433, opened 2022-08-09
PCOM-9743 Return card should support I18N translations

hklein-logic:
Added linked receipts label and transaction label so that values can be passed in


hklein-logic:
Dependency on 

Commerce
https://github.com/JumpMind/commerce/pull/7316

Issue #2451, opened 2022-08-22
Gift receipt icon is not displayed when item is marked as gift receipted


VodianytskyiVI:
Gift receipt icon is not displayed when item is marked as gift receipted
https://jira.ae.com/browse/PCOM-10064

PR #2452, opened 2022-08-22
Fix: Gift receipt icon is not displayed when item is marked as gift receipted.

VodianytskyiVI:
Fix issue when Gift receipt icon is not displayed when item is marked as gift receipted.
### Issues Fixed
https://github.com/JumpMind/openpos-framework/issues/2451


MaximAndreevich:
@chenson42 
item-card.component.html had CLRF line endings. Changed to LF by commit policy. 
item-card.component.scss has too. should it be changed to LF?

chenson42:
why do the html files look like they entire file changed?  that might be hard to merge upward.

VodianytskyiVI:
> why do the html files look like they entire file changed? that might be hard to merge upward.

looks like because it contained wrong line separators. I took a look on similar files and they had LF when these 2 files didn't have LF
That's mentioned in comment from @MaximAndreevich above as well

chenson42 (approved):

PR #2458, opened 2022-08-23
Fixed: #2457 - Add openpos config for SymDS externalId

### LINKED ISSUE #2457: Allow SymmetricDS externalId to be configured separately from the openpos installationId
Allow SymmetricDS externalId to be configured separately from the openpos installationId.

New parameter added:
 - openpos.externalId

gwilmer:
### Issues Fixed
#2457


openpos-symds/src/main/java/org/jumpmind/pos/symds/DefaultSymDSConfigurator.java
```diff
@@ -23,6 +23,10 @@ public class DefaultSymDSConfigurator implements ISymDSConfigurator {
     @Value("${openpos.installationId:'not set'}")
     String installationId;
 
+    @Value("${openpos.externalId:not set}")
```
chenson42, line 26: I think this should be openops.symmetric.externalId

chenson42 (approved):

PR #2464, opened 2022-08-26
Feature/ts 12973 sell item detail screen alignment

pulakesh3:
changing version number from snapshot to release version


pulakesh3:
abandon this PR

Issue #2495, opened 2022-09-07
Update Java Form and Openpos DynamicForms to enable state/data-driven form columns


brandon-lewis:
**Is your feature request related to a problem? Please describe.**
Not related to a technical problem, but I am noticing that we are building new screens dialogs for every different combination of form field columns and display orders. Building new screens like this requires Commerce and Openpos PRs, so this process is somewhat tedious and time consuming. This seems like an opportunity to add new column support for the dynamic forms to shorten development time as well prevent specific Form/FormFields screen bloat.

**Describe the solution you'd like**
I think the cleanest and most non-disruptive solution to this is to add a new field to the Java `Form.class` and Openpos `form.interface` files. There is a CSS style called `column-count` which will display all child elements in an equal number of columns by the CSS property value, wrapping vertically.

The DynamicFormUIMessage and its data trickles down to display one of two Angular components: `dynamic-form-component` or `dynamic-form-dialog-component`. By adding a new "numColumns" field that defaults to 1, we gain the ability to specify from the state how many columns we want the form to be displayed with. We _could_ also add this numColumns field to CTX_FORM to make it more data-driven. The dynamic-form-part component does not have a display style, so adding a default 1 column-count property to the form styling should not have any undesirable effects. No HTML/template changes necessary.

![image](https://user-images.githubusercontent.com/28220454/188918272-0284bab1-2dc4-4929-96eb-ae319266911c.png)

### Property Value Examples
Below are examples of the column-count property on the same form without any other template or styling changes.

**column-count: 1** (default)
![image](https://user-images.githubusercontent.com/28220454/188912644-264d05ae-3217-47ad-9e87-53c3e53ae475.png)

**column-count: 2** (notice that form fields are wrapped vertically, which is in alignment with the general practice of other forms across the POS)
![image](https://user-images.githubusercontent.com/28220454/188912684-0e58b688-09fe-4303-af67-a604cd950947.png)

**column-count: 3**
![image](https://user-images.githubusercontent.com/28220454/188912696-0fcf5ff4-7967-4b59-9ced-cedb864e8eb0.png)

**Describe alternatives you've considered**
We can continue building new openpos screens/components each time a client wants a different row/column order for specific forms.

PR #2496, opened 2022-09-07
Enable state/data-driven Form columns in DynamicForms

brandon-lewis:
Added a `numColumn` field to Form.java and form.interface to enable setting the column-count display setting from flow states. More context of the problem statement is detailed in the fixed issue (Feature Request).

If a `numColumn` greater than 1 is set on the Form object, the dynamic-form-part component will have its column-count will be set accordingly and the child form fields will be set to block display to appropriately wrap vertically (as column-count will not wrap inline elements). Otherwise, we default the column-count to 1 to maintain current look and feel.

Manually tested on desktop/tablet/mobile on many different POS screens and dialogs with 1 and 2+ column-counts.

### Issues Fixed
https://github.com/JumpMind/openpos-framework/issues/2495

## Testing
The dynamic-form-field component is used in a few places outside of dynamic-form. Here are screenshots of those components with the default column-count: 1 as proof that default Forms without a state-specified column-count don't look any different.

### Report Filters
image

image

### Transaction Search

image

image


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/dynamic-form-part/dynamic-form-part.component.html
```diff
@@ -2,12 +2,13 @@
 <app-form *ngIf="screenData" (submitFormEvent)="submitForm()" [form]="form" [validateOnFirstLoad]="screenData.validateOnFirstLoad">
     <div responsive-class class="form-part-body">
         <app-content-card>
-            <mat-card-content mat-dialog-content class="form-content" responsive-class>
-                <app-dynamic-form-field *ngFor="let field of screenData.formElements" [formGroup]="form" [formField]="field" (changed)="onFieldChanged($event)"></app-dynamic-form-field>
+            <mat-card-content mat-dialog-content class="form-content" [ngStyle]="{'column-count': screenData.numColumns > 1 ? screenData.numColumns : 1}" responsive-class>
```
jchirico8, line 5: add a `&& !isMobile` to this condition instead of leaning on `!important` to override this value in mobile and tablet views. I feel that's more clear as to why you wouldn't be seeing multiple columns if you were in an ipad resolution

brandon-lewis: @jchirico8 Updated to do proper isMobile checks and removed style overrides.

jchirico8 (approved):

PR #2548, opened 2022-09-30
Unable to add update the DML's in customization layer

### LINKED ISSUE #2549: Unable to add update the DML's in customization layer
**Describe the bug**
When we override a DML file, the commerce DML queries were not getting loaded in startup.

**Label the bug**
milestone

**To Reproduce**
1. Add a new query in the existing DML file in the customization layer.
2. When the server starts the commerce DML file was overridden and only customization DML queries were loaded.
3. Application was partially failing.

**Expected behavior**
When we override a DML file, both the commerce and the customization DML queries should get loaded at startup

**Screenshots**
NA

**Additional context**
https://petcoalm.atlassian.net/browse/PDPOS-9520

jayasree-litmus7:
We have introduced a new DML(sls-dml in the customization layer and due to a merge logic issue in openpos(DMLTemplates.java), when there is a second DML template, the merge logic fails and the final list will only contain DML statements from the last template file.
The merge logic works fine for query template merging. The same needs to be done for DmlTemplates as well.

closes #2549


openpos-persist/src/test/java/org/jumpmind/pos/persist/impl/DMLTemplatesTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+
+    @Before
+    public void setUp() {
+        dmlTemplates = new DmlTemplates();
+
+    }
+
+    @Test
+    public void addDmls() {
```
jrmihalick, line 26: Please name the test to be more explanatory of what is being tested and add some comments to explain it.

jrmihalick (changes requested):
Also do a PR for master once this is approved.  Thx.

openpos-persist/src/test/java/org/jumpmind/pos/persist/impl/DMLTemplatesTest.java
jayasree-litmus7: Pushed the changes as suggested. Please have a relook

jrmihalick (approved):

PR #2552, opened 2022-09-30
Server can reject new connections if old connection never disconnects…

chenson42:
Server can reject new connections if old connection never disconnects.  The original intent of this was to prevent multiple clients from connecting.  This PR takes a different approach by closing the old session and accepting the new session.


jchirico8 (approved):

chenson42:
we only allow one deviceid to attach to a server at a time.  the way this used to work was to track the device auth token that is connected and if another tries to connect then we reject it.   we think there is a case where the tracked connected session can be abandoned.   because of that we changed the approach to close prior sessions rather than reject new ones.   this will fix the scenario where we incorrectly had a reference to a previously connected device.   however, the result of this PR will cause two sessions with the same auth token to ping pong disconnecting.   Previously when you had the same url open in two tabs one of the sessions would not connect.  now they just ping pong back and forth disconnecting the other session.   i think we need to fix this eventually, but i could not find a good hook in the client to know that the server forceably disconnected the client.

chenson42:
https://jumpmind.atlassian.net/browse/JMC-43

PR #2554, opened 2022-09-30
Add Handling For European Currency Amounts 1000 Or More

JoeBrownlee:
URBN had display problems on their old version of the framework with European currency amounts 1000 or more. While Commerce replaces their comma decimal point with a period, if there is also a digit group separator that has a period, you end up with a number like 1.000.00, which isn't a valid number. This change eliminates the digit group separators so the number is valid. The code that handles this in `currency-text.component.ts` is unchanged since version 0.5. The back end is fine and handles everything correctly; only the display is wrong.

Note that it was nearly impossible to test this on the current version of Commerce. Setting Commerce to a location in France did not seem to change the number display. URBN's old OrPOS code uses the Java locale to display numbers correctly for the given location. Commerce may need some work to handle that. What I did was use the debugger to set the numeric text on the Product Inquiry Item Details page, which at least displays the euro symbol. Commerce generally displays a dollar sign everywhere. The debugger shows that internally, Commerce _knows_ the amounts are in Euros despite what is being displayed.

So for now, these changes will likely lurk in their respective components until Commerce can better handle displaying different number formats according to the locale.

Note that these changes should be a no-op for non-European number formats.

What I did to configure my environment:
- Set the `openpos.location.general.businessUnitLocale` parameter to `fr_FR`.
- Set the `openpos.location.general.defaultCountryId` parameter to `FR`.
- Set the `openpos.sell.defaultCurrencyId` parameter to `EUR`.
- Set the `openpos.orders.isoCurrencyCode` parameter to `"EUR"`.
- Add a Business Unit in `CTX_BUSINESS_UNIT` and devices in `DEV_DEVICE`. You can use the SQL below to add my rows to these tables.
- Personalize to the 'Euro Test' Business Unit and register device.

```
INSERT INTO PUBLIC.CTX_BUSINESS_UNIT (BUSINESS_UNIT_ID,GEO_CODE,ISO_CURRENCY_CODE,BUSINESS_UNIT_NAME,CREATE_TIME,CREATE_BY,LAST_UPDATE_TIME,LAST_UPDATE_BY,GOVERNMENT_ID) VALUES
	 ('06242','06242','EUR','Euro Test','2018-12-24 13:07:03','me','2018-12-24 13:07:03','you',NULL);
INSERT INTO PUBLIC.DEV_DEVICE (DEVICE_ID,APP_ID,PAIRED_DEVICE_ID,LOCALE,TIMEZONE_OFFSET,BUSINESS_UNIT_ID,DESCRIPTION,INSTALLATION_ID,CREATE_TIME,CREATE_BY,LAST_UPDATE_TIME,LAST_UPDATE_BY,TAG_COUNTRY,TAG_STATE,TAG_STORE_NUMBER,TAG_STORE_TYPE,TAG_APP_PROFILE,TAG_BRAND,PARENT_DEVICE_ID) VALUES
	 ('06242-001','pos',NULL,'fr_FR','0','06242','null','06242','2020-09-25 09:35:28','openpos-dev','2022-09-30 16:10:45.351','personalization','FR','*','*','*','*','*',NULL),
	 ('06242-001-pay','simPaymentDevice',NULL,'fr_FR','0','06242',NULL,'06242','2022-09-30 09:37:51.548','openpos-dev','2022-09-30 09:37:51.55','personalization','FR','*','*','*','*','*',NULL);
```

### Issues Fixed
This resolves [#7771](https://github.com/JumpMind/commerce/issues/7771).

Broader changes will be needed to make Commerce truly "Europe-friendly".


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/currency-text/currency-text.component.ts
```diff
@@ -84,6 +84,10 @@ export class CurrencyTextComponent implements OnChanges {
     }
 
     protected formatEuText(text: string): string {
+        if (text.match('[0-9]*\.[0-9]*,[0-9]*')) {
```
jchirico8, line 87: It's sparse, but this component has a spec file. Could you add some tests for this function to ensure it works as intended?

JoeBrownlee: OK, Joe, done, and good thing you suggested it. If the value wasn't in a range, it wasn't running the European number reformatting at all. Fixed that. I had copied the fix from v0.5, but that version only has single values -- no ranges. So in that version, the reformat is always checked for. On master, it was only done if part of a range.

jchirico8 (dismissed):
Looks like you have a lint error

jchirico8 (dismissed):

jrmihalick (approved):

PR #2566, opened 2022-10-11
changes done to replace tender icon with pad lock if additional icon …

chandankumarcentric:
https://petcoalm.atlassian.net/browse/PDPOS-9171

Commerce issue #7888

refer to commerce PR https://github.com/JumpMind/commerce/pull/7906

BEFORE:
image

After:
Desktop:
image

Ipad:
image

Mobile:
image


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/option-button/option-button.component.html
```diff
@@ -1,12 +1,13 @@
 <button responsive-class #button mat-flat-button [disabled]="disabled" (click)="clickFn()" [type]="inputType"
     color=primary-inverse class="option-item" tabindex="0">
 
-    <app-icon [iconName]="optionIcon" [iconClass]="optionSize" [class.muted-color]="disabled"
+    <app-icon *ngIf="!additionalIcon" [iconName]="optionIcon" [iconClass]="optionSize" [class.muted-color]="disabled"
```
jrmihalick, line 4: Leaving this attribute named as `additionalIcon` is going to be pretty confusing after this change. How about we rename to `iconOverride`?  @jchirico8, @stojankz, @chenson42 ... any opinions?

jchirico8: Is this change necessary even? If you're replacing the existing icon with a lock can't that be done server side on the tender option itself?

jrmihalick: @jchirico8 Yes, good point.  @chandankumarcentric : make it so.  😄

chandankumarcentric: i did the changes. I need help in handling this part in commerce PR:
https://github.com/JumpMind/commerce/pull/7906/files#diff-dd439fbbe655983250142be3c94021d75d29f735e5397b8780498a8725a1966c

chandankumarcentric: or should i keep the logic for additionalIcon and only handle TenderOptions case on server side? Please suggest.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/status/status-details/selector/peripheral-selector.component.html
```diff
@@ -16,7 +16,6 @@
         appArrowTabItem
         [attr.tabindex]="i"
         [optionTitle]="o.displayName"
-        [additionalIcon]="o.powerLevelIcon"
```
jchirico8, line 19: This powerLevelIcon is still required. I think if @JJStarrett used the additionalIcon in its existing location to show the payment device battery level, then you shouldn't change any of this

chandankumarcentric: > This powerLevelIcon is still required. I think if @JJStarrett used the additionalIcon in its existing location to show the payment device battery level, then you shouldn't change any of this

Thats what i was thinking to leave the additionalIcon as it is and handle only for TenderOption. Done please check now.

openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/option-button/option-button.component.html
chandankumarcentric: Done. Thanks.

openpos-flow/src/main/java/org/jumpmind/pos/core/ui/ActionItem.java
```diff
@@ -50,6 +50,8 @@ public class ActionItem implements Serializable {
     protected String imageUrl;
     @Builder.Default
     protected boolean enabled = true;
+    @Builder.Default
+    protected boolean iconMuted = false;
```
jchirico8, line 54: Primitive booleans default to false automatically. You don't need this or the builder.default
```suggestion
    protected boolean iconMuted;
```

chandankumarcentric: right. Thanks.

jrmihalick (approved):
Don't forget we need this in 3.3 also.

PR #2572, opened 2022-10-14
Feature/icon for store id and claim timeout adjust

### LINKED ISSUE #2570: Add icon to represent "Wrong Store"
**Is your feature request related to a problem? Please describe.**
Add an icon to represent "this device is not boarded for this store"

An icon to represent "this device is not boarded for this store"

**Describe alternatives you've considered**
We dismantle the entirety of the currency system, therefore eliminating the need for payment devices altogether and thus this feature. 

**Additional context**
AEO Jira
  - [10055](https://jira.ae.com/projects/PCOM/issues/PCOM-10055)

JJStarrett:
### Issues Fixed
 - Resolves #2570 
 - Resolves #2571 

### Summary
Adds a new icon to represent a payment device that is not "boarded for this store." Adyen gives us back a `storeId` when we do our diagnostics request, and that ID should match to our business unit. If it doesn't we'd like to disable it. This adds an icon to support that.

Additionally, the H400 seems to be hitting claim timeout exceptions in JavaPOS where the H300 seemed not to do so. I've bumped the timeout from 100ms to 200ms.


openpos-peripheral/src/main/java/org/jumpmind/pos/print/JavaPOSPrinter.java
```diff
@@ -45,11 +45,11 @@ public void open(String logicalName, EventCallbacks cb) throws JposException {
         try {
             this.printerName = logicalName;
             open((String) settings.get("printerName"));
-            claim(100);
+            claim(200);
```
chenson42, line 48: would it be wrong to make this 2000 or 1000?

JJStarrett: Bumped it to 2000. I've run with it at 60000 for debugging, so I think it should be fine

chenson42 (approved):

Issue #2573, opened 2022-10-14
H400 printer seems to take longer to claim than the H300 (4.2)


JJStarrett:
**Describe the bug**
The H400 printer will periodically fail to be claimed on back-to-back receipts with the 100ms claim timeout. 

**To Reproduce**
1. Print any sale receipt that normally prints both a sale and associate receipt
2. Observe that sometimes we fail with the JavaPOS `jpos.JposException: Claim timeout Error`

**Expected behavior**
The H300 seems to not have this problem, so we'd expect the H400 to function similarly. 

**Additional context**
AEO Jira
 - [11181](https://jira.ae.com/projects/PCOM/issues/PCOM-11181)

PR #2574, opened 2022-10-14
Adjust claim timeout slightly (4.2)

JJStarrett:
### Issues Fixed
 - #2573 

### Summary
Cherry-pick of part of https://github.com/JumpMind/openpos-framework/pull/2572, to bring the adjusted timeout to 4.2.


openpos-peripheral/src/main/java/org/jumpmind/pos/print/JavaPOSPrinter.java
```diff
@@ -45,11 +45,11 @@ public void open(String logicalName, EventCallbacks cb) throws JposException {
         try {
             this.printerName = logicalName;
             open((String) settings.get("printerName"));
-            claim(100);
+            claim(200);
```
chenson42, line 48: should it be even bigger than this?  what is wrong with having a timeout of 2 seconds?

JJStarrett: As far as I can tell no, it doesn't seem to cause any issues - I'll bump it up

chenson42 (approved):

PR #2578, opened 2022-10-18
Bugfix/pcom 10950 update all connected devices brand

MRUFFING67:
In reference of [AEO's PCOM-10950](https://jira.ae.com/browse/PCOM-10950).

Updating the brand would originally only affect the device that the brand was changed on.
Now, updating the brand will affect all connected devices. This change will cause the brand to automatically update on connected iPads/customer-displays as well.


openpos-flow/src/main/java/org/jumpmind/pos/core/service/DevToolsActionListener.java
```diff
@@ -250,19 +250,30 @@ private void setCustomerDisplayAuthData(String deviceId, Message message) {
...
-                    .filter(param -> "brandId".equals(param.getParamName()))
-                    .findFirst()
-                    .ifPresent(existingBrand -> existingBrand.setParamValue(brand));
-            devicesRepository.saveDevice(currentDevice);
-            stateManagerFactory.changeBrand(deviceId, brand);
-        } catch (DeviceNotFoundException ex) {
+    private void updateBrand(String brand) {
+
+        // Updates brand for ALL connected devices
+        Set<String> connectedDevices = devicesRepository.getConnectedDeviceIds();
```
jchirico8, line 256: This grabs every device connected to the server. It sounds fine when talking about fixed registers because it's just the register and 1 customer display (unless mobile devices have failed over to it), but the ISP will have several devices that will all be updated. I don't know if that's their intention

jchirico8 (commented):
I'm not sold that we should change the dev menu brand switch option to update anything besides the current device. This isn't meant to be used outside of test environments, so it shouldn't need to be updated to meet business requirements

openpos-flow/src/main/java/org/jumpmind/pos/core/service/DevToolsActionListener.java
MRUFFING67: Right, it would also update the associate display if the customer-display brand is updated from the customer display's DevTools

jchirico8: I mean that it'll update all 4-6 ipads that are connected to the stores mobile server regardless of pairing

PR #2592, opened 2022-10-26
update openpos-service.md

JJStarrett:
### Summary
Update openpos-service.md documentation on endpoint extension, overriding, practices.


openpos-service/src/main/docs/openpos-service.md
```diff
@@ -221,7 +300,7 @@ public class SearchCustomerEndpointOverride {
 }
 ~~~
 
-Oftentimes, like in this example, you will want to augment the existing Endpoint and not completely replace it.  We are recomending you **avoid Java inheritence** in the Endpoint overrides. Instead, you can use composition to cleanly refer to the base Endpoint.  Here is an example where the **base Endpoint is injected, invoked, and it's response can then be augmented by the override**:
+Oftentimes, like in this example, you will want to augment the existing Endpoint and not completely replace it.  We are recommending you **avoid Java inheritance** in the Endpoint overrides. Instead, you can use composition to cleanly refer to the base Endpoint.  Here is an example where the **base Endpoint is injected, invoked, and it's response can then be augmented by the override**:
```
jrmihalick, line 303: I wonder if this is old?  Why would we avoid java inheritance?  We use it often in endpoint overrides.  I know it used to not work, but inheritance has worked for several releases now.

openpos-service/src/main/docs/openpos-service.md
```diff
@@ -239,7 +318,47 @@ public class SearchCustomerEndpointOverride {
 }
 ~~~
 
-So here the key is to @Autowire in the base Endpoint, instead of intertwining the two Endpoint implementations through inheritence.
+So here the key is to `@Autowire` in the base Endpoint, instead of intertwining the two Endpoint implementations through inheritance
+
+## Repositories and Persistence
+
+Normally a module will declare a DBSessionFactory and DBSession in its module declaration class. This makes it natural to write one or more **repository** classes. Any logic that saves or retries data should be placed in a repository class.  The repository class should expose a clean API to the service which does not suggest what persistence technology it is using to save or retrive the data.
```
jrmihalick, line 325: retries?  Should that be retrieves?

JJStarrett: I deleted and re-added this, git's showing this as new. I avoided cleaning any spelling mistakes to not make it too confusing to see what I added, but I think that train's already left. I'll clean the whole doc real quick.

openpos-service/src/main/docs/openpos-service.md
JJStarrett: Well actually - @chenson42 should we still be avoiding inheritance here? I know I personally don't when implementing overridden endpoints, but before I remove this - is that something I _should_ be avoiding?

JJStarrett: Gotcha - went ahead and switched this line to prefer inheritance

chenson42 (approved):

PR #2600, opened 2022-10-28
added variable to store the last focussed index

chandankumarcentric:
fixed issue [#8014](https://github.com/JumpMind/commerce/issues/8014)

Commerce PR# https://github.com/JumpMind/commerce/pull/8015


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/actions/action.service.ts
```diff
@@ -17,6 +17,7 @@ export class ActionService implements OnDestroy {
     private blockActions: boolean;
     private actionPayloads: Map<string, () => void> = new Map<string, () => void>();
     private actionDisablers = new Map<string, BehaviorSubject<boolean>>();
+    private lastFocusedItemIndex = new BehaviorSubject<number>(-1);
```
jrmihalick, line 20: I think you are going to need to find another way to do this. It's out of place to track this in the ActionService.

jrmihalick (changes requested):

openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/actions/action.service.ts
jrmihalick: How is it being handled on other screens without this change?

chandankumarcentric:
no change needed

PR #2611, opened 2022-11-03
A race condition could give you the incorrect endpoint and because we cache Methods you can end up with mismatched implementations

chenson42:
A race condition could give you the incorrect endpoint and because we cache Methods you can end up with mismatched implementations.  This PR caches all endpoints by path and then by implementation.  It gets rid of the extra map of training endpoints because we now have a data structure that will support them.

This PR also removes @EndpointOverride.   Please use endpoint implementations instead.


jrmihalick:
giphy (2)

chenson42:
> giphy (2) [ giphy (2) ](https://user-images.githubusercontent.com/11838640/199839331-62cfb99e-59ca-4d0e-9042-9cfb72d98bcb.gif) [ ](https://user-images.githubusercontent.com/11838640/199839331-62cfb99e-59ca-4d0e-9042-9cfb72d98bcb.gif)

I take it I am Ernie.  I also noted that you did not approve the change.

jrmihalick:
I was just thinking that elimination of EndpointOverride and the ensuing configuration changes necessary will be a very noteworthy change for implementations in the future that migrate from versions prior to 5.0.  We were going to use tagging on issues, but I am wondering if someone should be keeping track of changes like this.

PR #2616, opened 2022-11-04
add ${}

MRUFFING67:
Fixing reference. Was originally referencing variable as if it was in the same file as reference


openpos-gradle/src/main/resources/java.gradle
```diff
@@ -197,7 +197,7 @@ task develop {
 }
 
 dependencies {
-    implementation group: 'org.yaml', name: 'snakeyaml', version: snakeYamlVersion
+    implementation group: 'org.yaml', name: 'snakeyaml', version: ${snakeYamlVersion}
```
chenson42, line 200: I don't think you need the ${}.   You only use that when evaluating a variable as part of a groovy string (in double quotes)

PR #2618, opened 2022-11-04
Bugfix/add displayproperty getter

sathyakonatham:
### Issues Fixed
Update DisplayProperty to have a getter to return the first element of the values list to have backward compatibility with 3.2 client  

Petco Jira Tickets : PDPOS-9925


openpos-flow/src/main/java/org/jumpmind/pos/core/model/DisplayProperty.java
```diff
@@ -42,4 +42,13 @@ public void addValue(String value) {
         }
         this.values.add(value);
     }
+
+    @Deprecated
+    public String getValue() {
+        if (!this.values.isEmpty()) {
```
jrmihalick, line 48: Best to use `CollectionUtils.isNotEmpty()` here

jrmihalick (changes requested):

jrmihalick (approved):

PR #2647, opened 2022-11-22
issue-2639 scss tweak to avoid additional text and icon overlap

### LINKED ISSUE #2639: Button component additional text and additional icon overlap
Button component displays additional text and additional icon on top of each other

image

1. Use a button component
2. Provide additional text and additional icon
3. Render the component

**Expected behavior**
Additional text and additional icon should display separately

image

steveswing:
### Issues Fixed

#2639 

Note this PR is closed without being merged because the overall tender button layout is not appealing. A work-around for Petco was to set the button's icon to the padlock (PASSWORD) to indicate the option requires manager override vs setting the additional icon when additional text might be defined as well. Without this fix both additional icon and additional text are rendered on top of each other.


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/components/option-button/option-button.component.scss
```diff
@@ -36,16 +36,13 @@
             display: grid;
             grid-template-columns: 25% auto 25%;
             grid-template-rows: auto;
-            grid-template-areas: ". line additional";
+            grid-template-areas: ". line icon additional";
```
steveswing, line 39: `icon` could go after `additional` instead if `grid-template-columns` values of `25% auto 25%` render things correctly.

PR #2651, opened 2022-11-22
Added removeExtension to AbstractModel

aakibladhani:
Added removeExtension to AbstractModel.


jrmihalick:
This is necessary because we have cases for backward compatibility where an extension may only exist in a current release of commerce, but that extension does not exist in a prior version of commerce.  If the prior version of commerce calls an endpoint in the current version and that current version returns an object instance of the extension, the extension object will fail to deserialize in the prior version of commerce.

jrmihalick (approved):

PR #2656, opened 2022-11-28
Fixing address form component

MRUFFING67:
Related to but not technically the bug in JMC-83: https://jumpmind.atlassian.net/browse/JMC-83

The form was poorly made (on the frontend) and all of the required fields weren't showing, so the next button wouldn't show as enabled (although you could still click it....)


openpos-client-libs/projects/openpos-client-core-lib/src/lib/shared/screen-parts/auto-complete-address-part/auto-complete-address-part.component.ts
```diff
@@ -124,25 +125,29 @@ export class AutoCompleteAddressPartComponent extends ScreenPartComponent<IForm>
     }
 
     private parseAddressField(element: IFormElement): boolean {
-        let isAddressField = false;
-        if (element.id === 'streetAddress') {
-            this.streetAddress = element;
-        } else if (element.id === 'addressLine2') {
-            this.addressLine2 = element;
-            isAddressField = true;
-        } else if (element.id === 'locality') {
```
chenson42, line 133: Looks like locality was intentional.  How do we know we are not breaking someone else?

MRUFFING67: @chenson42 This component is used is 3 parent components, none of which are used themselves

chenson42 (approved):

Issue #672, opened 2020-04-23
Investigate if ITypeCodes can be deserialized from JSON using the datatype of the attribute being deserialized into


jrmihalick:
As I was staring at the implementation of the `ITypeCodeDeserializer#deserialize` method, it made me wonder if the Jackson provided DeserializationContext would provide the necessary typing information needed in order to deserialize an ITypeCode.  If so, this could eliminate the need to store the ITypeCode class information in serialized Json. See `DeserializationContext.getContextualType()` method as a starting point.

I'm not sure why I didn't think of this originally. Or maybe I did and there was a reason it wouldn't work.

PR #2660, opened 2022-11-30
Allow ITypeCode deserializer to work without class specified

chenson42:
When passing type codes back and forth between a smart ui it would be nice to not have to specify the class.   This PR allows ITypeCode deserializer to work without class specified.  If the class is not specified then it uses the class from the context (target).


stojankz (dismissed):

chenson42:
I'm merging, but @jrmihalick could you still take a peak?

jrmihalick (approved):
Looks like this will close #672.

stojankz (approved):

PR #2684, opened 2022-12-21
Screen does not refresh after a server restart on master

chenson42:
Screen does not refresh after a server restart on master


jchirico8 (approved):

jrmihalick:
Awesome ... I have been seeing this on 3.x for a while.  Thanks @chenson42

PR #2692, opened 2022-12-29
Don't send entire object graph to server for logging.  Limit sending to configured log level log level

chenson42:
Don't send entire object graph to server for logging.  Limit sending to configured log level log level


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/logging/console-scraper.service.ts
```diff
@@ -113,30 +113,14 @@ export class ConsoleScraper {
...
-
     // Stringifys given message and handles any circular object references,
     // preventing the JSON.stringify circular reference error
     private _jsonStringifySanitize(message: any): string {
         let cleansed = message;
 
         if (message && typeof message === 'object') {
             try {
-                cleansed = JSON.stringify(message, this._makeCircularRefReplacer());
+                cleansed = JSON.stringify(message, function (k, v) { return k && v && typeof v !== "number" ? (Array.isArray(v) ? "[object Array]" : "" + v) : v; });
```
jrmihalick, line 123: A little concerned about removing the circular ref logic here since we needed it before apparently

chenson42: it is no longer needed.  i'm only logging the first level of the object with this new function.

chenson42:
A customer was getting OOM errors on the server because of the size of the messages being sent across.   No need to lot the entire graph of an object that we'll never look at.

jrmihalick (dismissed):

PR #2718, opened 2023-02-01
Make area code read-only

rfmajor:
Expected result - While creating RR or while adding return customer entry field in MX, in Phone No field the +52 is for display only. It should not be editable. 

image

Changes were done in loyalty-customer-form-dialog files to comply with the above requirements. Now the result is:

image

Note: as of now this field is specific to MX, it's not displayed in other countries unless explicitly configured


gwilmer (changes requested):
It's not areaCode, it's countryCode...

gwilmer:
I don't think we should do this at all.  In the US, we don't display, nor store the country code.  We shouldn't do that in other countries either...  If we do, it needs to probably be a country code field on PhoneModel.  Rejecting this PR, and we will discuss what we want to do here...

PR #2719, opened 2023-02-02
JMC-173 - When showing two dialogs back to back where the first has auto focus and shows the ios keyboard, when the keyboard is dismissed a blank white area is left where the keyboard was.

chenson42:
### Summary
When showing two dialogs back to back where the first has auto focus and shows the ios keyboard, when the keyboard is dismissed a blank white area is left where the keyboard was.
### Issues Fixed
https://jumpmind.atlassian.net/browse/JMC-173


openpos-client-libs/projects/openpos-client-core-lib/src/lib/core/services/dialog.service.ts
```diff
@@ -148,10 +150,18 @@ export class DialogService {
...
                 forceReopen = dialog.dialogProperties.forceReopen;
             }
-            // By default we want to not allow the user to close by clicking off
+            // By default, we want to not allow the user to close by clicking off
             // By default we need the dialog to grab focus so you cannont execute actions on the screen
             // behind by hitting enter
             const dialogProperties: OpenPOSDialogConfig = { disableClose: !closeable, autoFocus: true };
+
+            if (dialog.screenType.startsWith('Prompt') && Capacitor.isNativePlatform()) {
+                console.info('[DialogService] Enhancing ' + dialog.screenType + 'to be positioned at the top')
```
brandon-lewis, line 159: Looks like a space is missing between the dialog.screenType and the rest of the log message. Otherwise, it all looks fitting to me.

